PR #1, opened 2022-09-06
Implement WordPress core management: install/config

weiiwang01:


src/charm.py
```diff
@@ -740,6 +721,139 @@ def _gen_wp_config(self):
...
+    def _wp_is_installed(self):
+        """Check if WordPress is installed (check if WordPress related tables exist in database)"""
+        process = self._container().exec(["wp", "core", "is-installed"])
+        try:
+            process.wait()
+            return True
+        except ops.pebble.ExecError:
+            return False
+
+    def _wp_install_cmd(self):
```
jdkandersson, line 744: DO you think a docstring might be helpful here?

weiiwang01: Fixed in the new commit.

src/charm.py
```diff
@@ -635,13 +610,19 @@ def _wp_generate_password(length=64):
             characters = string.ascii_letters + "!@#$%^&*()" + "-_ []{}<>~`+=,.;:/?|"
             return "".join(secrets.choice(characters) for _ in range(length))
 
-        return {
+        wp_secrets = {
             field: _wp_generate_password()
             for field in self._wordpress_secret_key_fields()
         }
+        wp_secrets["default_admin_password"] = secrets.token_urlsafe(32)
```
jdkandersson, line 617: For my understanding, the `default_admin_password` secret seems to be generated differently to the other secrets using `secrets.token_urlsafe` instead of the `_wordpress_secret_key_fields` function. What is the reason for that please?

Also, it looks like `_wordpress_secret_key_fields` doesn't generate the `default_admin_password` key, should it? Or is it more appropriate to define it specially here?

src/charm.py
```diff
@@ -635,13 +610,19 @@ def _wp_generate_password(length=64):
...
+        return wp_secrets
 
     def _replica_relation_data(self):
```
jdkandersson, line 620: Do you think a docstring would be helpful here?

src/charm.py
```diff
@@ -635,13 +610,19 @@ def _wp_generate_password(length=64):
...
-        return self.model.get_relation("wordpress_replica").data[self.app]
+        relation = self.model.get_relation("wordpress-replica")
+        if relation is None:
+            return {}
```
jdkandersson, line 623: Looking through how the return value of `_replica_relation_data` is used, it seems that in many cases the keys that are expected are the same as the secret keys generated by the `_wordpress_secret_key_fields` function, wouldn't returning an empty dictionary here just lead to `KeyError` in all those functions? Maybe raising an error here is more appropriate?

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
         )
         return "\n".join(wp_config)
 
+    def _container(self):
+        return self.unit.get_container("wordpress")
```
jdkandersson, line 725: Do you think a docstring here and in any other functions being added in the PR would be helpful?

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+        This operation is idempotence.
+        """
+        if self._wordpress_service_exists():
+            self._container().stop("wordpress")
+
+    def _wp_is_installed(self):
+        """Check if WordPress is installed (check if WordPress related tables exist in database)"""
+        process = self._container().exec(["wp", "core", "is-installed"])
+        try:
+            process.wait()
```
jdkandersson, line 742: Do you think a timeout would be beneficial here? Not sure if it is possible/ already enabled by default

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+            return True
+        except ops.pebble.ExecError:
+            return False
+
+    def _wp_install_cmd(self):
+        """Generate wp-cli command used to install WordPress on database"""
+        initial_settings = yaml.safe_load(self.model.config["initial_settings"])
+        admin_user = initial_settings.get("user_name", "admin_username")
```
jdkandersson, line 750: It looks like the default settings define a value for `user_name` and `admin_email` in the `config.yaml` file, should the same values be used here? Potentially even just read the values from that file so that they are only defined in one place?

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+        admin_email = initial_settings.get("admin_email", "name@example.com")
+        default_admin_password = self._replica_relation_data()["default_admin_password"]
+        admin_password = initial_settings.get("admin_password", default_admin_password)
+        return [
+            "wp",
+            "core",
+            "install",
+            "--url=localhost",
+            "--title=The {} Blog".format(self.model.config["blog_hostname"] or self.app.name),
```
jdkandersson, line 759: Would an f-string work here and be easier to read? The same for the other `.format` calls below

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+            "--admin_user={}".format(admin_user),
+            "--admin_email={}".format(admin_email),
+            "--admin_password={}".format(admin_password)
+        ]
+
+    def _wp_install(self):
+        """Install WordPress (create WordPress required tables in DB)"""
+        process = self._container().exec(self._wp_install_cmd())
+        process.wait()
```
jdkandersson, line 768: Timeout here as well?

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+
+    def _start_server(self):
+        """Start WordPress (apache) server. On leader unit, also make sure WordPress is installed
+
+        Check if the pebble layer has been added, then check the installation status of WordPress,
+        finally start the server. The installation process only run on the leader unit. This
+        operation is idempotence.
+        """
+        self._init_pebble_layer()
+        if self.unit.is_leader():
```
jdkandersson, line 793: Probably my limited understanding, is wordpress only installed on the leader? Also, shouldn't the application container come with wordpress installed already?

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+            if not self._wp_is_installed():
+                self._wp_install()
+        self._container().start("wordpress")
+
+    @staticmethod
+    def _wp_config_path():
```
jdkandersson, line 799: Does this have to be a function or could it just be, e.g., a class variable?

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+
+        It will fail under the following two circumstances:
+          - Peer relation data not ready
+          - Config doesn't provide valid database information and db relation hasn't
+            been established
+
+        It will check if the current wp-config.php file matches the desired config.
+        If not, update the wp-config.php file.
+
+        It will also check if WordPress is installed (WordPress-related tables exist in db).
```
jdkandersson, line 825: Just wondering why part of it is in brackets here?

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+        If not, install WordPress (create WordPress required tables in db).
+
+        If any update is needed, it will stop the apache server first to prevent any requests
+        during the update for security reasons.
+        """
+        if not self._replica_consensus_reached():
+            self.unit.status = WaitingStatus("Waiting for unit consensus")
```
jdkandersson, line 832: Is this the best place to set the status of the charm? Could it, for example, be done by whatever function calls this function based on the return value of this function? Would that make the charm easier to maintain?

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+            self._stop_server()
```
jdkandersson, line 833: Similar to above, should the caller make decisions about whether to stop the container versus this function?

src/charm.py
```diff
@@ -740,6 +721,144 @@ def _gen_wp_config(self):
...
+            self._push_wp_config(wp_config)
+        self._start_server()
+        return True
+
+    def _reconciliation(self, _event):
+        if not self._container().can_connect():
+            self.unit.status = WaitingStatus("Waiting for pebble")
+            return
+        if not self._core_reconciliation():
```
jdkandersson, line 859: At least for me, it would be much easier to understand this if the `_core_reconcilliation` function would just return whether the core is ready potentially with a reason why it is or not and this function then updates the status and starts or stops the server based on the output rather than mixing setting the status in this function and the `_core_reconiclliation` function

src/charm.pyi
```diff
@@ -23,4 +24,31 @@ class WordpressCharm(ops.charm.CharmBase):
...
 
-    def _gen_wp_config(self) -> str: ...
\ No newline at end of file
+    def _gen_wp_config(self) -> str: ...
+
+    def _container(self) -> ops.charm.model.Container: ...
+
+    def _wordpress_service_exists(self) -> bool: ...
+
+    def _stop_server(self): ...
```
jdkandersson, line 33: Should the return value here be `None`? Same comment applies to any other functions here without a return type

tests/unit/test_charm_2.py
```diff
@@ -1,12 +1,48 @@
+import os
```
jdkandersson, line 1: Would it be possible to come up with a more descriptive name for the file than `test_charm_2`? E.g., instead of `2` maybe something about how it is different or what specifically it tests?

tests/unit/test_charm_2.py
```diff
@@ -1,12 +1,48 @@
...
 import unittest
 import unittest.mock
 
 import ops.testing
 from charm import WordpressCharm
 
+METADATA_FILE = os.path.join(os.path.split(__file__)[0], "..", "..", "metadata.yaml")
```
jdkandersson, line 8: What about `Pathlib`: https://docs.python.org/3/library/pathlib.html, it is a higher level API which I find very easy to read and understand

tests/unit/test_charm_2.py
```diff
@@ -1,12 +1,48 @@
...
+
 
 class TestWordpressK8s(unittest.TestCase):
     def setUp(self):
+        container_fs = {}
```
jdkandersson, line 13: Is this variable necessary? Could this and the below line be combined?

tests/unit/test_charm_2.py
```diff
@@ -1,12 +1,48 @@
...
+        def mock_wp_is_installed(_self):
+            db_host = get_current_db_host(_self)
+            return db_host in installed_database_hosts
+
+        def mock_wp_install(_self):
+            db_host = get_current_db_host(_self)
+            installed_database_hosts.add(db_host)
+
+        self.container_patch = unittest.mock.patch.multiple(
+            WordpressCharm,
```
jdkandersson, line 39: Any particular reason for not using the https://docs.python.org/3/library/unittest.mock.html library?

tests/unit/test_charm_2.py
```diff
@@ -29,6 +66,12 @@ def test_generate_wp_secret_keys(self):
         """
         self.harness.begin()
         secrets = self.harness.charm._generate_wp_secret_keys()
+        self.assertIn(
+            "default_admin_password",
+            secrets,
+            "wordpress should generate a default admin password"
+        )
+        del secrets["default_admin_password"]
```
jdkandersson, line 74: For my understanding, why is this being deleted?

tests/unit/test_charm_2.py
```diff
@@ -215,3 +258,103 @@ def in_same_line(content, *matches):
                 ),
                 "db info in config should takes precedence over the db relation"
             )
+
+    def test_wp_install_cmd(self):
+        """
```
jdkandersson, line 263: What about arrange?

tests/unit/test_charm_2.py
```diff
@@ -215,3 +258,103 @@ def in_same_line(content, *matches):
...
+        act: generate wp-cli command to install WordPress
+        assert: generated command should match current config and status
+        """
+        consensus = self._setup_replica_consensus()
```
jdkandersson, line 267: Would it be worth separating the arrange and assert blocks using an empty line to make this test easier to read and differentiate between executing behaviour and checking results? Similar comment for the rest of the tests

tests/unit/test_charm_2.py
```diff
@@ -215,3 +258,103 @@ def in_same_line(content, *matches):
...
+            """
+        })
+        install_cmd = self.harness.charm._wp_install_cmd()
+        self.assertIn("--admin_user=test_admin_username", install_cmd)
+        self.assertIn("--admin_email=test@test.com", install_cmd)
+        self.assertIn("--admin_password=test_admin_password", install_cmd)
+
+    def test_core_reconciliation_before_peer_relation_ready(self):
+        """
+        arrange: before peer relation established
```
jdkandersson, line 293: Any other pre-conditions? E.g., the charm has been created?

jdkandersson (changes requested):

tests/unit/test_charm_2.py
weiiwang01: Updated, thanks!

tests/unit/test_charm_2.py
weiiwang01: I will add some empty lines between arrange and act/assert, thanks!

tests/unit/test_charm_2.py
weiiwang01: There's no arrange here, it works without any precondition.

tests/unit/test_charm_2.py
weiiwang01: Because I wrote the following test statements first, I don't want to rewrite this entire test case when adding tests for default_admin_password :D

tests/unit/test_charm_2.py
weiiwang01: I think this is `unitest.mock`.

tests/unit/test_charm_2.py
weiiwang01: Just don't want to access `self` variable in function closure, I guess it works, but it will look confusing.

tests/unit/test_charm_2.py
weiiwang01: Yes, definitely `Pathlib`, but this line here is currently not used, so I will just remove this for now, thanks!

tests/unit/test_charm_2.py
weiiwang01: The 2 here is because I don't want to change the original `test_charm.py` since it has some conflicts between the old code and new code. I will remove the `test_charm.py` and rename `test_charm_2.py` to `test_charm.py` in final stage.

src/charm.pyi
weiiwang01: I think they are none by default? But I will add them for less confusion.

src/charm.py
weiiwang01: each `_*_reconciliation` function is part of the `_reconciliation` that changes the state of application (on/off, configurations) to match the state of charm config / relations. The idea is `_reconciliation` delegates to `_*_reconciliation` and treat them as a blackbox. The `_core_reconciliation` is a little bit different here since it will create some essential configurations that other reconciliation function like `_plugin_reconciliation`, `_theme_reconciliation` needs, so it pass some information (`True`/ `False`) outside. I think I will change this to not passing anything outside and add more checks to other  reconciliation functions. It might cost more computer/io resources but after that all the reconciliation is a complete blackbox.

src/charm.py
weiiwang01: See the comment below.

src/charm.py
weiiwang01: See the comment below.

src/charm.py
weiiwang01: The words in brackets explains "installed", WordPress installation is a little different than the traditional install like `apt install` or `npm install`.

src/charm.py
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: "WordPress install" is just creating tables in mysql database, so it will only need to do once for a single site.

src/charm.py
weiiwang01: Added, thanks!

src/charm.py
weiiwang01: Yes, the f-string was introduced in Python 3.6, I think it can be used, I will start update them all, thanks!

src/charm.py
weiiwang01: The default value for `initial_settings` is `"user_name: admin\nadmin_email: devnull@example.com"`. If `initial_setting` being set to `""`, it will be equivalent to `"user_name: admin_username\nadmin_email: name@example.com"`. It's confusing here, but for the sake of backward compatibility, just implement as is.

src/charm.py
weiiwang01: Added, thanks!

src/charm.py
weiiwang01: Added, thanks!

src/charm.py
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: Added, thanks!

src/charm.py
weiiwang01: `default_admin_password` is for humans and others are only for WordPress to use internally so they are a little bit different. All secrets listed in `_wordpress_secret_key_fields` is generate so that is perfectly compatible with the PHP function `wp_salt`. The reason `default_admin_password` doesn't use `wp_salt` is because `wp_salt` contains too much special characters that would be too difficult for human to read.

jdkandersson: Makes sense, thanks

src/charm.py
jdkandersson: Does the charm currently use `admin_username` and `name@example.com`?

src/charm.py
jdkandersson: Ah I see, maybe the function name can be changed to be more obvious? I read it as some sort of software install or something like that

src/charm.py
jdkandersson: Ah it clarifies what is meant by install, I see

tests/unit/test_charm_2.py
jdkandersson: Thats even better!

tests/unit/test_charm_2.py
jdkandersson: Oh I didn't notice that the functions below also make use of it, never mind

tests/unit/test_charm_2.py
jdkandersson: Yes, I was more meaning, for example, using a mock.MagicMock instead of defining a functions like `mock_wp_is_installed`, although I see a bit of custom logic that might be difficult to implement using a MagicMock

src/charm.py
weiiwang01: if user set the `initial_setting` config to `""` explicitly, then it will use `admin_username` and `name@example.com`

tests/unit/test_charm_2.py
jdkandersson: Ah, it might help if the `arrange: no pre-condition` line is there or something like that just to ensure the reader know you haven't forgotten it 😉

tests/unit/test_charm_2.py
weiiwang01: Yes, this is a simulation of container file system and process execution, so it's more complex than MagicMock

src/charm.py
jdkandersson: Ok

tests/unit/test_charm_2.py
jdkandersson: 👍

tests/unit/test_charm_2.py
weiiwang01: Gotcha, updated

tests/unit/test_charm_2.py
jdkandersson: 👍

src/charm.py
jdkandersson: 👍

src/charm.py
weiiwang01: I think just follow the WordPress terminology for now, the underlining wp-cli command is `wp core install` so we better match that.

jdkandersson: Ah I see, makes sense

src/charm.py
jdkandersson: I'm guessing the docstring is on its way, otherwise I'm good with this

jdkandersson (approved):

jdkandersson:
I'll change the minimum number of reviewers to 2 in line with the contributing guide

weiiwang01:
> Weird that it can be merged with just 1 approval?

I think because the target branch `sidecar` is not a protected branch, only `main` is.

jdkandersson:
> > Weird that it can be merged with just 1 approval?
> 
> I think because the target branch `sidecar` is not a protected branch, only `main` is.

I see, it still might be worth waiting for 1 more review anyway, pretend the sidecar branch is main 😉

src/charm.py
```diff
@@ -740,6 +730,153 @@ def _gen_wp_config(self):
         )
         return "\n".join(wp_config)
 
+    def _container(self):
+        """Get the WordPress workload container"""
+        return self.unit.get_container("wordpress")
```
arturo-seijas, line 735: It might be useful defining both the container name and the service name in a constant

src/charm.py
```diff
@@ -740,6 +730,153 @@ def _gen_wp_config(self):
...
+            "description": "WordPress server",
+            "services": {
+                "wordpress": {
+                    "override": "replace",
+                    "summary": "WordPress server (apache)",
+                    "command": "apache2ctl -D FOREGROUND",
+                }
+            },
+        }
+        if not self._wordpress_service_exists():
```
arturo-seijas, line 792: I don't think you need this condition. Adding the layer independently of the previous status should do it

src/charm.py
```diff
@@ -740,6 +730,153 @@ def _gen_wp_config(self):
...
+        if self._wordpress_service_exists():
+            self._container().stop("wordpress")
+
+    def _wp_is_installed(self):
+        """Check if WordPress is installed (check if WordPress related tables exist in database)"""
+        process = self._container().exec(["wp", "core", "is-installed"], timeout=60)
+        try:
+            process.wait()
+            return True
+        except ops.pebble.ExecError:
```
arturo-seijas, line 754: Some logging here might come in handy

src/charm.py
```diff
@@ -740,6 +730,153 @@ def _gen_wp_config(self):
...
+        self._init_pebble_layer()
+        if self.unit.is_leader():
+            if not self._wp_is_installed():
+                self._wp_install()
+            if self._current_wp_config() is None:
+                # For security reasons, never start WordPress server if wp-config.php not exists
+                raise FileNotFoundError(
+                    "required file (wp-config.php) for starting WordPress server not exists"
+                )
+        self._container().start("wordpress")
```
arturo-seijas, line 811: I'd go for self._container.pebble.replan_services() instead: https://github.com/canonical/operator/blob/main/ops/pebble.py#LC1625

arturo-seijas (changes requested):

src/charm.py
weiiwang01: Fixed. thanks!

src/charm.py
weiiwang01: Fixed, thanks!

src/charm.py
weiiwang01: I think `replan_services` will only start the services if the services definition changes. Here the definition never changes after been set, but server will start and stop multiple times while running.

arturo-seijas (approved):
LGTM

PR #3, opened 2022-09-14
Add WordPress functionality test

weiiwang01:
Add WordPress functionality integration test to test the following basic feature of the WordPress instance on charm:
- Login as the admin user
- Create a new post
- Add comments to a blog post

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -128,4 +128,12 @@ async def fixture_get_unit_ip_list(
             ip_list.append(units[key].address)
         return ip_list
 
-    return _get_unit_ip_list
+    yield _get_unit_ip_list
+
+
+@pytest_asyncio.fixture(scope="function", name="unit_ip_list")
+async def fixture_unit_ip_list(
+        get_unit_ip_list
+):
+    """A fixture containing ip addresses of current units"""
+    yield await get_unit_ip_list()
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -99,6 +99,7 @@ async def fixture_get_unit_status_msg_list(ops_test: pytest_operator.plugin.OpsT
 def fixture_application_name():
     return "wordpress"
 
+
 @pytest_asyncio.fixture(scope="function", name="default_admin_password")
 async def fixture_default_admin_password(ops_test: pytest_operator.plugin.OpsTest):
     return_code, stdout, stderr = await ops_test.juju(
@@ -116,4 +117,5 @@ async def fixture_default_admin_password(ops_test: pytest_operator.plugin.OpsTes
     assert (
             result["status"] == "completed"
     ), "get-initial-password action should have a status of completed"
+
     yield result["results"]["password"]
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,5 +1,3 @@
-import json
-
 import pytest
 import juju.action
 import pytest_asyncio
@@ -24,76 +22,6 @@ async def fixture_app_config(request, ops_test: pytest_operator.plugin.OpsTest):
 
     await application.set_config(original_config)
     await ops_test.model.wait_for_idle()
-
-
-@pytest_asyncio.fixture(scope="module", name="get_app_status")
-async def fixture_get_app_status(ops_test: pytest_operator.plugin.OpsTest):
-    """Helper function to get the status of application
-
-    Returns a async function that can retrieve the current status of a application, if application
-    name not given, default to the WordPress application. The status is in string form.
-    """
-
-    async def _get_app_status(application_name: str = "wordpress"):
-        status = await ops_test.model.get_status()
-        return status.applications[application_name].status.status
-
-    return _get_app_status
-
-
-@pytest_asyncio.fixture(scope="module", name="get_app_status_msg")
-async def fixture_get_app_status_msg(ops_test: pytest_operator.plugin.OpsTest):
-    """Helper function to get the status message of application
-
-    Similar to fixture_get_app_status, but return status message instead
-    """
-
-    async def _get_app_status(application_name: str = "wordpress"):
-        status = await ops_test.model.get_status()
-        return status.applications[application_name].info
-
-    return _get_app_status
-
-
-@pytest_asyncio.fixture(scope="module", name="get_unit_status_list")
-async def fixture_get_unit_status_list(ops_test: pytest_operator.plugin.OpsTest):
-    """Helper function to get status of units
-
-    Returns a async function that can retrieve the current status of all the units from
-    an application, if application name not given, default to the WordPress application.
-    The status is in string form. The result list is sorted by unit id.
-    """
-
-    async def _get_unit_status_list(application_name: str = "wordpress"):
-        status = await ops_test.model.get_status()
-        result = []
-        units_dict = status.applications[application_name].units
-        unit_names = sorted(units_dict.keys(), key=lambda n: int(n.split("/")[-1]))
-        for unit_name in unit_names:
-            result.append(units_dict[unit_name].workload_status.status)
-        return result
-
-    return _get_unit_status_list
-
-
-@pytest_asyncio.fixture(scope="module", name="get_unit_status_msg_list")
-async def fixture_get_unit_status_msg_list(ops_test: pytest_operator.plugin.OpsTest):
-    """Helper function to get status messages of units
-
-    Similar to fixture_get_unit_status_list, but return status message instead
-    """
-
-    async def _get_unit_status_msg_list(application_name: str = "wordpress"):
-        status = await ops_test.model.get_status()
-        result = []
-        units_dict = status.applications[application_name].units
-
-        unit_names = sorted(units_dict.keys(), key=lambda n: int(n.split("/")[-1]))
-        for unit_name in unit_names:
-            result.append(units_dict[unit_name].workload_status.info)
-        return result
-
-    return _get_unit_status_msg_list
 
 
 @pytest.fixture(scope="module", name="application_name")
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,3 +1,5 @@
+import json
+
 import pytest
 import pytest_asyncio
 import juju.application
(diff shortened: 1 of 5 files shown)


tests/integration/conftest.py
```diff
@@ -96,3 +98,22 @@ async def _get_unit_status_msg_list(application_name: str = "wordpress"):
...
+        "json"
+    )
+    assert (
+            return_code == 0 and not stderr
+    ), "juju run-action get-initial-password should not fail"
+    result = json.loads(stdout)["unit-wordpress-0"]
+    assert (
+            result["status"] == "completed"
+    ), "get-initial-password action should have a status of completed"
+    yield result["results"]["password"]
```
arturo-seijas, line 119: Missing blank line here

arturo-seijas (commented):
LGTM, just one picky comment

tests/integration/conftest.py
weiiwang01: Fixed, thanks!

arturo-seijas (approved):
LGTM

tests/integration/conftest.py
```diff
@@ -96,3 +98,24 @@ async def _get_unit_status_msg_list(application_name: str = "wordpress"):
 @pytest.fixture(scope="module", name="application_name")
 def fixture_application_name():
     return "wordpress"
+
+
+@pytest_asyncio.fixture(scope="function", name="default_admin_password")
+async def fixture_default_admin_password(ops_test: pytest_operator.plugin.OpsTest):
+    return_code, stdout, stderr = await ops_test.juju(
```
jdkandersson, line 105: There is a native way of doing this: https://pythonlibjuju.readthedocs.io/en/latest/api/juju.unit.html#juju.unit.Unit.run_action

The unit should be accessible under the `<wordpress application>.units[0]`

tests/integration/test_core.py
```diff
@@ -89,9 +91,64 @@ async def test_mysql_relation(
...
+
+
+@pytest.mark.asyncio
+async def test_get_initial_password_action(default_admin_password):
+    """
+    arrange: after WordPress charm has been deployed
+    act: run get-initial-password action
+    assert: get-initial-password action should return default admin password
+    """
+    assert (
```
jdkandersson, line 110: Do you think the action should be triggered from within the test?

tests/integration/test_core.py
```diff
@@ -89,9 +91,64 @@ async def test_mysql_relation(
...
+            len(default_admin_password) > 8
+    ), "get-initial-password action should return the default password"
+
+
+@pytest.mark.asyncio
+async def test_wordpress_functionality(
```
jdkandersson, line 116: Do you think it is worth making this test name more descriptive, e.g., include login, post, comment?

tests/integration/test_core.py
```diff
@@ -89,9 +91,64 @@ async def test_mysql_relation(
...
+        post_content = secrets.token_hex(16)
+        post = wp_client.create_post(
+            title=post_title,
+            content=post_content,
+        )
+        homepage = wp_client.get_homepage()
+        assert (
+                post_title in homepage and post_content in homepage
+        ), "admin user should be able to create a new post"
+        comment = secrets.token_hex(16)
```
jdkandersson, line 145: Do you think it is worth putting a comment here explaining why a secret token is used as the comment? I'm guessing the test just needs some random string? Would some fixed string or a string including e.g., the name of the unit work?

tests/integration/test_core.py
```diff
@@ -89,9 +91,64 @@ async def test_mysql_relation(
...
+):
+    """
+    arrange: after WordPress charm has been deployed and db relation established
+    act: test WordPress basic functionality (login, post, comment)
+    assert: WordPress works normally as a blog site
+    """
+    units = (await ops_test.model.get_status()).applications[application_name].units.values()
+    for unit in units:
+        unit_ip = unit.address
+        wp_client = WordpressClient(
```
jdkandersson, line 129: Do you think this would be better for a fixture to make the test shorter and easier to understand/ read?

tests/integration/wordpress_client.py
```diff
@@ -0,0 +1,122 @@
...
+        self.password = password
+        self._session = requests.session()
+        self.timeout = 10
+        self._login()
+        try:
+            self._session.get(
+                f"http://{self.host}/wp-json/",
+                timeout=self.timeout
+            ).json()
+        except requests.exceptions.JSONDecodeError:
```
jdkandersson, line 22: For my understanding, could you explain what is going on here please?

tests/integration/wordpress_client.py
```diff
@@ -0,0 +1,122 @@
...
+            if not is_admin:
+                raise ValueError("set options-permalink manually or login with an admin account")
+            self._set_options_permalink()
+
+    def _login(self):
```
jdkandersson, line 27: Do you think docstrings would be worthwhile here? Same comment applies for the rest of the functions in this module

tests/integration/wordpress_client.py
```diff
@@ -0,0 +1,122 @@
...
+            f"http://{self.host}/wp-json/wp/v2/posts/",
+            json={
+                "status": "publish",
+                "title": title,
+                "content": content
+            },
+            headers={"X-WP-Nonce": nonce},
+            timeout=self.timeout
+        )
+        assert (
```
jdkandersson, line 87: Should the asserts be here or potentially in the test code? E.g., could this be done as an exception? Applies to other assert statements here as well

tests/integration/wordpress_client.py
```diff
@@ -0,0 +1,122 @@
...
+                response.status_code == 200
+        ), "admin user should able to set WordPress options-permalink to /%postname%/"
+
+    def create_post(self, title: str, content: str):
+        new_post_page = self._session.get(
+            f"http://{self.host}/wp-admin/post-new.php",
+            timeout=self.timeout
+        )
+        new_post_page = new_post_page.text
+        nonce = json.loads(re.findall('var wpApiSettings = ([^;]+);', new_post_page)[0])["nonce"]
```
jdkandersson, line 76: For all the different lines calculating the nonce, how stable do you think this way of getting the nonce is? Are the pages likely to change and thus the way the nonce is captured in the tests need changing too?

tests/unit/test_charm_2.py
```diff
@@ -393,3 +393,43 @@ def test_core_reconciliation(self):
...
+        event_mock.fail = unittest.mock.MagicMock()
+        return event_mock
+
+    def test_get_initial_password_action_before_replica_consensus(self):
+        """
+        arrange: before peer relation established but after charm created
+        act: run get-initial-password action
+        assert: get-initial-password action should fail
+        """
+        self.harness.begin_with_initial_hooks()
```
jdkandersson, line 410: Is it worth separating the act and assert here? Similar comment applies to the other tests

jdkandersson (changes requested):
Thanks for this

tests/integration/conftest.py
weiiwang01: Fixed, thanks!

tests/integration/test_core.py
weiiwang01: Currently this test case only includes login, post, and comment, but I am planning to expand that with more functionality test like create user, non admin user, upload image, post with image etc. A more detailed name would be too long in that case.

tests/integration/test_core.py
weiiwang01: It's just a nonce, this test will run multiple times and each time needs a unique title and content. But I think a comment about this is a good ideal.

tests/integration/wordpress_client.py
weiiwang01: WordPress will not expose the wp-json endpoints by default, here, the client check if wp-json is exposed, if not change the permalink setting to expose that.

tests/integration/wordpress_client.py
weiiwang01: Fixed, thanks!

tests/integration/wordpress_client.py
weiiwang01: Yes, it might change in the future, the more proper way is to run some PHP code in the WordPress instance, but that would be too complex for our purpose here. I think it's acceptable to update this following WordPress upgrades.

tests/integration/wordpress_client.py
weiiwang01: If using the exception, it will lose the pytest integration, so things like actual vs except comparison will be gone.

tests/unit/test_charm_2.py
weiiwang01: Fixed, thanks!

tests/integration/test_core.py
weiiwang01: There are some things about scaling a container based charm I am not sure yet. Will update on this later, thanks!

tests/integration/test_core.py
weiiwang01: Fixed, thanks!

tests/integration/wordpress_client.py
jdkandersson: Ah that makes sense, do you think it is worth adding a comment to explain that? I was quite confused before and that explanation helped a lot 😄

tests/integration/test_core.py
weiiwang01: I think I will move this WordPress test procedure to a class method in `WordpressClient` because I believe this test procedure will run in a lot of different test cases.

tests/integration/wordpress_client.py
jdkandersson: E.g., `create_post_assert_201_response`

merkata (approved):
Looks really great!

tests/integration/wordpress_client.py
weiiwang01: Move the test procedures to a class method in `WordpressClient`

tests/integration/wordpress_client.py
weiiwang01: Added, thanks.

tests/integration/test_core.py
weiiwang01: Fixed, thanks!

tests/integration/conftest.py
```diff
@@ -101,21 +102,30 @@ def fixture_application_name():
...
+
+    async def _get_unit_ip_list():
+        status = await ops_test.model.get_status()
+        units = status.applications[application_name].units
+        ip_list = []
+        for key in sorted(units.keys(), key=lambda n: int(n.split("/")[-1])):
+            ip_list.append(units[key].address)
+        return ip_list
+
+    return _get_unit_ip_list
```
jdkandersson, line 131: Nice work moving this to a fixture, could this fixture just return the value of the `_get_unit_ip_list` rather than returning the function?

weiiwang01: In some of the future test cases, the unit replicas number could change in the test case (for testing charm behavior after add-unit), but I think I can add another fixture named just `unit_ip_list` for getting the IP address directly for test cases without unit number changing.

jdkandersson (approved):

PR #4, opened 2022-09-16
Implement WordPress theme management

weiiwang01:
Implementing WordPress theme management logic:
- Add a new config option `themes`.
- Add `_theme_reconciliation` function to install / uninstall themes according to `themes` setting.
- Add unit tests and integration test to WordPress theme management system.


weiiwang01:
> Hey @weiiwang01 , this is nice work! I'm seeing it is a draft PR - can you comment on what is available and what should be added, so far everything looks good. Thanks!

Updated the PR description with integration tests. Converting this PR to "Ready for review".

Thank you!

merkata (approved):
It looks really good and I'm approving. I see that the repo doesn't require approvals (the team defaults to two or more approvals on a PR), so that can be set in the settings imo. And we need to think about when to enable GH Actions, I assume tests are running locally?

config.yaml
```diff
@@ -107,3 +107,9 @@ options:
       YAML dictionary with keys named after WordPress settings and the desired values.
       Please note that the settings will be reset to values provided every time hooks run.
     default: ""
+  themes:
+    type: string
+    description: |
+      Theme slugs of themes to be installed, separated by comma. Not including default
+      installed themes.
+    default: ""
```
arturo-seijas, line 115: Missing blank line here

src/charm.py
```diff
@@ -954,6 +976,70 @@ def _core_reconciliation(self):
...
+    def _wp_theme_list(self):
+        """List all installed WordPress themes"""
+        process = self._run_wp_cli(["wp", "theme", "list", "--format=json"], timeout=600)
+        if process.return_code != 0:
+            raise RuntimeError("wp theme list command failed: " + process.stderr)
+        try:
+            return json.loads(process.stdout)
+        except json.decoder.JSONDecodeError:
+            raise RuntimeError(
+                "wp theme list command failed, stdout is not json: " + repr(process.stdout)
```
arturo-seijas, line 988: Should we use an f-string instead?

src/charm.py
```diff
@@ -954,6 +976,70 @@ def _core_reconciliation(self):
...
+        logger.debug("Start theme reconciliation process")
+        current_installed_themes = set(t["name"] for t in self._wp_theme_list())
+        logger.debug("Currently installed themes %s", current_installed_themes)
+        themes_in_config = [t.strip() for t in self.model.config["themes"].split(",")]
+        if themes_in_config == [""]:
+            themes_in_config = []
+        desired_themes = set(
+            itertools.chain(
+                themes_in_config,
+                self._WORDPRESS_DEFAULT_THEMES
```
arturo-seijas, line 1026: Are we using the same linting settings across charms? I would have expected a ',' in here

arturo-seijas (changes requested):

src/charm.py
weiiwang01: Yes, but I heard about the f-string vulnerability recently so I don't know what to use.

src/charm.py
weiiwang01: Linting is disabled currently due to the legacy code and new code are mixed together right now. I will address this issue in the later stage.

weiiwang01:
> It looks really good and I'm approving. I see that the repo doesn't require approvals (the team defaults to two or more approvals on a PR), so that can be set in the settings imo. And we need to think about when to enable GH Actions, I assume tests are running locally?

This PR doesn't require approving because the target branch is `sidecar` and that's not a protected branch.
Yes, all tests are running locally and I agree adding GitHub actions, but GitHub actions can be only effective after they being added to main branch. The current main branch is the old master branch from launchpad. Since I want to keep the main branch a working version of the charm, I will add GitHub actions when sidecar branch is mature enough.

config.yaml
jdkandersson, line 115: Can we have a new line at the end of this file?

src/charm.py
```diff
@@ -744,20 +770,25 @@ def _stop_server(self):
         if self._wordpress_service_exists():
             self._container().stop(self._SERVICE_NAME)
 
-    def _wp_is_installed(self):
-        """Check if WordPress is installed (check if WordPress related tables exist in database)"""
-        logger.debug("Check if WordPress is installed")
+    def _run_wp_cli(self, cmd, timeout=60, combine_stderr=False):
+        Result = collections.namedtuple("WordpressCliExecResult", "return_code stdout stderr")
```
jdkandersson, line 774: Should this be `Result = collections.namedtuple("WordpressCliExecResult", ["return_code", "stdout stderr"])`?

src/charm.py
```diff
@@ -744,20 +770,25 @@ def _stop_server(self):
...
+            timeout=timeout)
         try:
-            process.wait()
-            return True
-        except ops.pebble.ExecError:
-            return False
+            stdout, stderr = process.wait_output()
+            return Result(0, stdout, stderr)
+        except ops.pebble.ExecError as e:
+            return Result(e.exit_code, e.stdout, e.stderr)
```
jdkandersson, line 786: Nice! It works for both the happy and error case

src/charm.py
jdkandersson: f-strings are only an issue in the case of log messages? f-strings should be fine here. Unless you are thinking of another vulnerability?

src/charm.py
```diff
@@ -954,6 +976,70 @@ def _core_reconciliation(self):
...
+        """Install WordPress theme"""
+        process = self._run_wp_cli(
+            # --force will overwrite any installed version of the theme,
+            # without prompting for confirmation
+            ["wp", "theme", "install", theme, "--force"],
+            timeout=600,
+            combine_stderr=True
+        )
+        if process.return_code != 0:
+            raise RuntimeError("wp theme list command failed: " + process.stdout)
```
jdkandersson, line 1001: f-string?

src/charm.py
```diff
@@ -954,6 +976,70 @@ def _core_reconciliation(self):
...
+
+    def _wp_theme_delete(self, theme):
+        """Uninstall WordPress theme"""
+        process = self._run_wp_cli(
+            ["wp", "theme", "delete", theme, "--force"],
```
jdkandersson, line 1006: Should this always use `--force`? Are there any risks with using force to uninstall a theme?

src/charm.py
```diff
@@ -954,6 +976,70 @@ def _core_reconciliation(self):
...
+
+    def _theme_reconciliation(self):
+        """Reconciliation process for WordPress themes
+
+        Install and uninstall themes to match the themes setting in config
+        """
+        logger.debug("Start theme reconciliation process")
+        current_installed_themes = set(t["name"] for t in self._wp_theme_list())
+        logger.debug("Currently installed themes %s", current_installed_themes)
+        themes_in_config = [t.strip() for t in self.model.config["themes"].split(",")]
```
jdkandersson, line 1020: What about  `themes_in_config = [t.strip() for t in self.model.config["themes"].split(",")] if self.model.config["themes"].strip() else []`? Would mean the following `if` statement isn't necessary

src/charm.py
```diff
@@ -954,6 +976,70 @@ def _core_reconciliation(self):
...
+        if themes_in_config == [""]:
+            themes_in_config = []
+        desired_themes = set(
+            itertools.chain(
+                themes_in_config,
+                self._WORDPRESS_DEFAULT_THEMES
+            )
+        )
+        install_themes = desired_themes - current_installed_themes
```
jdkandersson, line 1029: Nice use of `set`!

src/charm.py
```diff
@@ -954,6 +976,70 @@ def _core_reconciliation(self):
...
+        uninstall_themes = current_installed_themes - desired_themes
+        for theme in install_themes:
+            try:
+                logger.debug("Install theme: %s", repr(theme))
```
jdkandersson, line 1033: What type os theme here? Is it something other than a `string`?

tests/integration/test_core.py
```diff
@@ -92,31 +93,6 @@ async def test_mysql_relation(
     )
 
 
-@pytest.mark.asyncio
```
jdkandersson, line 95: Should this test be removed? Is it no longer necessary?

tests/integration/test_core.py
```diff
@@ -133,3 +109,78 @@ async def test_wordpress_functionality(
...
+
+@pytest.mark.asyncio
+async def test_wordpress_default_themes(
+        unit_ip_list,
+        get_theme_list_from_ip
+):
+    """
+    arrange: after WordPress charm has been deployed and db relation established
+    act: check installed WordPress themes
+    assert: All default themes should be installed
```
jdkandersson, line 122: Very picky, should `All` be lower case here?

tests/integration/test_core.py
```diff
@@ -133,3 +109,78 @@ async def test_wordpress_functionality(
...
+    """
+    for unit_ip in unit_ip_list:
+        assert (
+                set(WordpressCharm._WORDPRESS_DEFAULT_THEMES) ==
+                set(get_theme_list_from_ip(unit_ip))
+        ), "default themes installed should match default themes defined in WordpressCharm"
+
+
+@pytest.mark.asyncio
+async def test_wordpress_install_uninstall_themes(
```
jdkandersson, line 132: Would it be worth also having a test where a config change results in both installing a theme and uninstalling a theme? What about installing multiple themes? Could potentially be a parametrised version of this test

tests/integration/test_core.py
```diff
@@ -133,3 +109,78 @@ async def test_wordpress_functionality(
...
+                theme not in get_theme_list_from_ip(unit_ip)
+        ), "theme should be uninstalled after removing from the themes setting in config"
+
+
+@pytest.mark.asyncio
+async def test_wordpress_theme_installation_error(
+        ops_test: pytest_operator.plugin.OpsTest,
+        application_name
+):
+    invalid_theme = "sgkeahrgalejr"
```
jdkandersson, line 167: what about calling it some version of `invalid-theme-sgkeahrgalejr`? It can help with readability if it ends up in any log messages etc if it has at least a portion of it be meaningful

tests/unit/test_charm_2.py
```diff
@@ -435,3 +455,50 @@ def test_get_initial_password_action(self):
...
+        })
+
+        self.assertEqual(
+            self.installed_themes,
+            set(self.harness.charm._WORDPRESS_DEFAULT_THEMES),
+            "installed themes should match the default installed themes "
+            "with the default themes config"
+        )
+
+        # Currently, Ops test framework has a bug preventing starting an already
```
jdkandersson, line 480: Excellent documentation for this! Do you think a date might help as well just to trigger a possible review if it has been a while?

jdkandersson (changes requested):

config.yaml
weiiwang01: Fixed, thanks!

src/charm.py
weiiwang01: Fixed, thanks!

config.yaml
weiiwang01: Fixed, thanks!

src/charm.py
weiiwang01: It works in both way https://docs.python.org/3/library/collections.html#collections.namedtuple
```python3
import collections

collections.namedtuple("Test", "a b")(a=1, b=2)
> Test(a=1, b=2)
```

src/charm.py
weiiwang01: Fixed, thanks!

src/charm.py
weiiwang01: I think it will cause some trouble when uninstall a currently activated theme without "--force", but the theme activation is not managed by juju, so force remove the theme here.

src/charm.py
weiiwang01: I came up with a even better way to do this
```
themes_in_config = [
    t.strip() for t in self.model.config["themes"].split(",")
    if t.strip()
]
```

src/charm.py
weiiwang01: No it's string, using `repr` here so it can be interpreted more clearly even the string contains whitespace or other special characters.

tests/integration/test_core.py
weiiwang01: This test case is move to a different location in the same file.

tests/integration/test_core.py
weiiwang01: Fixed, thanks

tests/integration/test_core.py
weiiwang01: Fixed, thanks!

tests/integration/test_core.py
weiiwang01: Fixed, thanks!

tests/unit/test_charm_2.py
weiiwang01: I will take a look at this problem this week to confirm if this problem exists and maybe submit a issue or PR for this.

src/charm.py
jdkandersson: Ah, makes sense!

tests/integration/test_core.py
jdkandersson: Ah

jdkandersson (approved):

arturo-seijas (approved):
LGTM

PR #7, opened 2022-09-29
WordPress plugin management (activation/option)

weiiwang01:
- Add control login to manage activation and options of three charm-managed plugins (akismet, openid, openstack-objectstorage-k8s)
- Add integration tests for openstack object storage plugin.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1192,7 +1192,7 @@ class WordpressCharm(CharmBase):
         return self._wrapped_run_wp_cli(["wp", "option", "delete", option])
 
     def _wp_plugin_activate(self, plugin):
-        """Active a WordPress plugin"""
+        """Activate a WordPress plugin"""
         logger.info("activate plugin %s", repr(plugin))
         return self._wrapped_run_wp_cli(["wp", "plugin", "activate", plugin])
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -715,7 +715,11 @@ class WordpressCharm(CharmBase):
             return relation.data[self.app]
 
     def _replica_consensus_reached(self):
-        """Test if the synchronized data required for WordPress replication are initialized."""
+        """Test if the synchronized data required for WordPress replication are initialized.
+
+        Returns:
+            True if the initialization of synchronized data has finished, else False.
+        """
         fields = self._wordpress_secret_key_fields()
         try:
             replica_data = self._replica_relation_data()
@@ -729,6 +733,12 @@ class WordpressCharm(CharmBase):
         Only the leader can update the data shared with all replicas. Leader should check if
         the data exist when leadership is established, generate required data and set it in
         the peer relation if not.
+
+        Args:
+            event: required by ops framework, not used.
+
+        Returns:
+            None.
         """
         if not self._replica_consensus_reached() and self.unit.is_leader():
             replica_relation_data = self._replica_relation_data()
@@ -741,6 +751,13 @@ class WordpressCharm(CharmBase):
 
         This method will set all db relation related states ``relation_db_*`` when db relation
         changes and will reset all that to ``None`` after db relation is broken.
+
+        Args:
+            event: An instance of opslib.mysql.MySQLDatabaseChangedEvent represents the new
+                database connection information.
+
+        Returns:
+            None.
         """
         self.state.relation_db_host = event.host
         self.state.relation_db_name = event.database
@@ -752,6 +769,9 @@ class WordpressCharm(CharmBase):
 
         This method will not check the validity of the configuration or current state,
         unless they are security related, in that case, an exception will be raised.
+
+        Returns:
+            The content of wp-config.php file in string.
         """
         wp_config = [
             textwrap.dedent("""\
@@ -807,16 +827,26 @@ class WordpressCharm(CharmBase):
         return "\n".join(wp_config)
 
     def _container(self):
-        """Get the WordPress workload container"""
+        """Get the WordPress workload container.
+
+        Returns:
+            The pebble instance of the WordPress container.
+        """
         return self.unit.get_container(self._CONTAINER_NAME)
 
     def _wordpress_service_exists(self):
+        """Check if the WordPress pebble layer exists.
+
+        Returns:
+            True if WordPress layer already exists, else False.
+        """
         return self._SERVICE_NAME in self._container().get_plan().services
 
     def _stop_server(self):
-        """Stop WordPress (apache) server
+        """Stop WordPress (apache) server, this operation is idempotence.
 
-        This operation is idempotence.
+        Returns:
+            None.
         """
         logger.debug("Ensure WordPress (apache) server is down")
         if (
@@ -834,6 +864,23 @@ class WordpressCharm(CharmBase):
             combine_stderr=False,
             timeout=60
     ):
+        """Execute a command in WordPress container.
+
+        Args:
+            cmd (List[str]): The command to be executed.
+            user (str): Username to run this command as, use root when not provided.
+            group (str): Name of the group to run this command as, use root when not provided.
+            working_dir (str):  Working dir to run this command in, use home dir if not provided.
+            combine_stderr (bool): Redirect stderr to stdout, when enabled, stderr in the result
+                will always be empty.
+            timeout (int): Set a timeout for the running program in seconds. Default is 60 seconds.
+                ``TimeoutError`` will be raised if timeout exceeded.
+
+        Returns:
+            A named tuple with three fields: return code, stdout and stderr. Stdout and stderr are
+            both string.
+        """
+
         Result = collections.namedtuple("CommandExecResult", "return_code stdout stderr")
         process = self._container().exec(
             cmd,
@@ -862,6 +909,11 @@ class WordpressCharm(CharmBase):
         return result
 
     def _run_wp_cli(self, cmd, timeout=60, combine_stderr=False):
+        """Execute a wp-cli command, this is a wrapper around :meth:`charm.WordpressCharm._run_cli`
+
+        See :meth:`charm.WordpressCharm._run_cli` for documentation of the arguments and return
+        value.
+        """
         result = self._run_cli(
             cmd,
             user=self._WORDPRESS_USER,
@@ -878,6 +930,18 @@ class WordpressCharm(CharmBase):
         Stdout and stderr are discarded, the result field of ExecResult is always none. The
         execution is considered success if return code is 0. The message field will be generated
         automatically based on command if ``error_message`` is not provided.
+
+        Args:
+            cmd (List[str]): The command to be executed.
+            timeout (int): Set a timeout for the running program, in seconds. Default is 60 seconds.
+                ``TimeoutError`` will be raised if timeout exceeded.
+            error_message (str) message in the return result if the command failed, if None, a default
+                error message will be provided in the result.
+
+        Returns:
+            A named tuple with three fields: success, result and message. ``success`` will be True
+            if the command succeed. ``result`` will always be None and ``message`` represents the
+            error message, in case of success, it will be empty.
         """
         result = self._run_wp_cli(cmd=cmd, timeout=timeout, combine_stderr=True)
         if result.return_code != 0:
@@ -890,7 +954,11 @@ class WordpressCharm(CharmBase):
             return self._ExecResult(success=True, result=None, message="")
 
     def _wp_is_installed(self):
-        """Check if WordPress is installed (check if WordPress related tables exist in database)"""
+        """Check if WordPress is installed (check if WordPress related tables exist in database)
+
+        Returns:
+            True if WordPress is installed in the current connected database.
+        """
         logger.debug("Check if WordPress is installed")
         return self._run_wp_cli(["wp", "core", "is-installed"]).return_code == 0
 
@@ -899,6 +967,10 @@ class WordpressCharm(CharmBase):
 
         Database info in config takes precedence over database info provided by relations.
         Return value is a dict containing four keys (DB_HOST, DB_NAME, DB_USER, DB_PASSWORD)
+
+        Returns:
+            A dict containing four keys "db_host", "db_name", "db_user" and "db_password". All
+            values are string.
         """
         database_info = {
             key.upper(): self.model.config[key] for key in
@@ -914,8 +986,9 @@ class WordpressCharm(CharmBase):
     def _test_database_connectivity(self):
         """Test the connectivity of the current database config/relation
 
-        Return a tuple of connectivity as bool and error message as str, error message will be
-        an empty string if charm can connect to the database.
+        Returns:
+            A tuple of connectivity as bool and error message as str, error message will be
+            an empty string if charm can connect to the database.
         """
         db_info = self._current_effective_db_info()
         try:
@@ -935,7 +1008,11 @@ class WordpressCharm(CharmBase):
             return False, f"MySQL error {err.errno}"
 
     def _wp_install_cmd(self):
-        """Generate wp-cli command used to install WordPress on database"""
+        """Generate wp-cli command used to install WordPress on database
+
+        Returns:
+            Wp-cli WordPress install command, a list of strings.
+        """
         initial_settings = yaml.safe_load(self.model.config["initial_settings"])
         admin_user = initial_settings.get("user_name", "admin_username")
         admin_email = initial_settings.get("admin_email", "name@example.com")
@@ -962,6 +1039,7 @@ class WordpressCharm(CharmBase):
             raise exceptions.WordPressInstallError("check logs for more information")
 
     def _init_pebble_layer(self):
+        """Ensure WordPress layer exists in pebble"""
         logger.debug("Ensure WordPress layer exists in pebble")
         layer = {
             "summary": "WordPress layer",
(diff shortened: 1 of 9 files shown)


src/charm.py
```diff
@@ -840,17 +848,32 @@ def _run_wp_cli(self, cmd, timeout=60, combine_stderr=False):
             result = Result(0, stdout, stderr)
         except ops.pebble.ExecError as e:
             result = Result(e.exit_code, e.stdout, e.stderr)
+        return_code = result.return_code
         if combine_stderr:
-            logger.debug("Run wp-cli command: %s\noutput: %s", cmd, result.stdout)
-        else:
             logger.debug(
-                "Run wp-cli command: %s\nstdout: %s\nstderr:%s",
+                f"Run command: %s\noutput: %s, return code {return_code}",
```
mthaddon, line 854: We shouldn't be using f-strings in logging output. See the contributing guide for some reasoning on why.

Same comment applies for a number of other entries here.

src/charm.py
```diff
@@ -1146,20 +1169,274 @@ def _addon_reconciliation(self, addon_type):
...
+            raise exceptions.WordPressBlockedStatusException(
+                f"Unable to config openid plugin, {result.message}"
+            )
+
+    def _apache_config_is_enabled(self, conf_name):
+        """Check if a specified apache configuration file is enabled"""
+        enabled_config = [
+            name for name in self._container().list_files("/etc/apache2/conf-enabled")
+        ]
+        return f"{conf_name}.conf" in enabled_config
```
mthaddon, line 1353: Missing an empty line here.

src/charm.py
```diff
@@ -1146,20 +1169,274 @@ def _addon_reconciliation(self, addon_type):
...
+    def _apache_enable_config(self, conf_name, conf):
+        """enables a specified configuration file within the apache2 configuration"""
```
mthaddon, line 1355: Can we capitalise the first word in the docstring?

src/charm.py
```diff
@@ -1146,20 +1169,274 @@ def _addon_reconciliation(self, addon_type):
...
+        self._stop_server()
+        self._container().push(
+            path=f"/etc/apache2/conf-available/{conf_name}.conf", source=conf
+        )
+        self._run_cli(["a2enconf", conf_name])
+        self._start_server()
+
+    def _apache_disable_config(self, conf_name):
+        """disables a specified configuration file within the apache2 configuration"""
```
mthaddon, line 1364: Can we capitalise the first word in the docstring?

mthaddon (commented):
I haven't done a full review here, but a few things seemed worth mentioning. Will take a closer look tomorrow, but have run out of day for now...

src/charm.py
jdkandersson: Please use `%s` and parameters in logging, even if we believe that it should be safe this open up an attack vector, applies for the other logging calls as well

src/charm.py
```diff
@@ -1146,20 +1169,274 @@ def _addon_reconciliation(self, addon_type):
...
+
+    def _wp_option_delete(self, option):
+        """Delete a WordPress option
+
+        It's not an error to delete a non-existent option (it's a warning though)
+        """
+        return self._wrapped_run_wp_cli(["wp", "option", "delete", option])
+
+    def _wp_plugin_activate(self, plugin):
+        """Active a WordPress plugin"""
```
jdkandersson, line 1195: Activate instead of active?

src/charm.py
```diff
@@ -1146,23 +1167,277 @@ def _addon_reconciliation(self, addon_type):
...
+
     def _theme_reconciliation(self):
         """Reconciliation process for WordPress themes
 
         Install and uninstall themes to match the themes setting in config
         """
         self._addon_reconciliation("theme")
 
+    def _wp_option_update(self, option, value, format="plaintext"):
+        """Create or update a WordPress option value
```
jdkandersson, line 1179: (Applies to all docstrings) we have a new standard for docstrings which includes descriptions of parameters and return values: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#docstrings

src/charm.py
weiiwang01: Updated all f-string in logging except some variables with fixed values.

src/charm.py
weiiwang01: Fixed, thanks!

src/charm.py
weiiwang01: Fixed, thanks!

src/charm.py
weiiwang01: Fixed, thanks!

src/charm.py
```diff
@@ -1146,23 +1167,279 @@ def _addon_reconciliation(self, addon_type):
...
+        current_plugins_activate_status = {
+            p["name"]: p["status"] for p in current_plugins
+        }
+        if plugin not in current_plugins_activate_status:
+            return self._ExecResult(
+                success=False,
+                result=None,
+                message=f"{action} a non-existent plugin {plugin}"
+            )
+        is_active = current_plugins_activate_status[plugin] == "active"
```
jdkandersson, line 1231: Do you think it is worthwhile to add a few new lines in this function to indicate a new set of logic? might make this more readable

src/charm.py
```diff
@@ -1146,23 +1167,279 @@ def _addon_reconciliation(self, addon_type):
...
+        akismet_key = self.model.config["wp_plugin_akismet_key"].strip()
+        if not akismet_key:
+            result = self._deactivate_plugin("akismet", [
+                "akismet_strictness",
+                "akismet_show_user_comments_approved",
+                "wordpress_api_key"
+            ])
+        else:
+            result = self._activate_plugin("akismet", {
+                "akismet_strictness": "0",
```
jdkandersson, line 1292: For these two values it seems they are hard coded? Will all users want this to be configured like this?

src/charm.py
```diff
@@ -1146,23 +1167,279 @@ def _addon_reconciliation(self, addon_type):
...
+    def _encode_openid_team_map(team_map):
+        """Convert wp_plugin_openid_team_map setting to WordPress openid_teams_trust_list option
+
+        example input: site-sysadmins=administrator,site-editors=editor,site-executives=editor
+        """
+        team_map_lines = []
+        i = 0
+        team_map_lines.append("a:{}:{{".format(len(team_map.split(","))))
+        for mapping in team_map.split(","):
+            i = i + 1
```
jdkandersson, line 1311: You can automatically get indexes if you wrap an iterable in the builtn `enumerate` function which returns the index and item as a tuple: https://docs.python.org/3/library/functions.html#enumerate

src/charm.py
```diff
@@ -1174,7 +1451,7 @@ def _reconciliation(self, _event):
             logger.info("Reconciliation process finished successfully.")
             self.unit.status = ops.model.ActiveStatus()
         except exceptions.WordPressStatusException as status_exception:
-            logger.info(f"Reconciliation process terminated early, reason: {status_exception}")
+            logger.info("Reconciliation process terminated early, reason: %s", status_exception)
             self.unit.status = status_exception.status
```
jdkandersson, line 1457: This file is getting quite long, at what point do you think this should be broken up into multiple files?

tests/integration/conftest.py
```diff
@@ -102,3 +106,18 @@ def _get_plugin_list_from_ip(unit_ip):
...
     return _get_plugin_list_from_ip
+
+
+@pytest.fixture
+def openstack_environment(request):
+    """Parse the openstack rc style configuration file from the --openstack-rc argument
+
+    Return a dictionary of environment variables and values.
+    """
+    rc_file = request.config.getoption("--openstack-rc")
```
jdkandersson, line 117: What if the parameter was not passed? Wouldn't `rc_file` be `None` in that case?

tests/integration/requirements.txt
```diff
@@ -1,5 +1,8 @@
 juju
 pytest
+Pillow
```
jdkandersson, line 3: Is this needed? I couldn't see it get imported anywhere? Maybe it is the `PIL` package?

tests/integration/requirements.txt
```diff
@@ -1,5 +1,8 @@
...
 requests
 pytest-cov
-pytest-operator
\ No newline at end of file
+pytest-operator
+python-swiftclient
+python-keystoneclient
```
jdkandersson, line 8: Is this needed? I couldn't see it get imported anywhere?

tests/integration/test_core.py
```diff
@@ -255,3 +265,103 @@ async def test_wordpress_plugin_installation_error(
...
+):
+    """
+    arrange: after WordPress charm has been deployed, db relation established and openstack swift
+        server ready
+    act: update charm configuration for openstack object storage plugin
+    assert: openstack object storage plugin should be installed after the config update and
+        WordPress openstack swift object storage integration should be set up properly.
+    """
+
+    swift_conn = swiftclient.Connection(
```
jdkandersson, line 286: Is this worth being in a fixture?

tests/unit/test_charm_2.py
```diff
@@ -630,3 +698,101 @@ def test_plugin_reconciliation(self):
...
+        """
+        self._standard_plugin_test(
+            plugin="openid",
+            plugin_config={
+                "wp_plugin_openid_team_map":
+                    "site-sysadmins=administrator,site-editors=editor,site-executives=editor"
+            },
+            excepted_options={
+                'openid_required_for_registration': '1',
+                'openid_teams_trust_list': 'a:3:{i:1;O:8:"stdClass":4:{s:2:"id";i:1;s:4:"team";s:14:"site-sysadmins";s:4:"role";s:13:"administrator";s:6:"server";s:1:"0";}i:2;O:8:"stdClass":4:{s:2:"id";i:2;s:4:"team";s:12:"site-editors";s:4:"role";s:6:"editor";s:6:"server";s:1:"0";}i:3;O:8:"stdClass":4:{s:2:"id";i:3;s:4:"team";s:15:"site-executives";s:4:"role";s:6:"editor";s:6:"server";s:1:"0";}}'
```
jdkandersson, line 776: Out of curiosity, how was this derived?

jdkandersson (changes requested):

src/charm.py
weiiwang01: Fixed in latter commit.

src/charm.py
weiiwang01: Fixed, thanks!

src/charm.py
weiiwang01: It's hard coded in the current stable version, so just follow that for now.

src/charm.py
weiiwang01: Yes, this is copied from the current stable version directly, this function is actually not quite right (encoding a PHP array in Python), I will replace this method in the future.

src/charm.py
weiiwang01: The upper half of this file is old code that's not needed anymore, I will remove them in a future PR so it will cut the file length in half.

tests/integration/conftest.py
weiiwang01: A new commit add a default and some comments on this option, it's an error to not pass --openstack-rc in integration tests, it's okay for unit tests.

tests/integration/requirements.txt
weiiwang01: It's used in integration tests (tests/integration/test_core.py)

tests/integration/requirements.txt
weiiwang01: `python-keystoneclient` is required for `python-swiftclient` to use Identity API v3.

tests/integration/test_core.py
weiiwang01: By design, when openstack swift integration is setup, there's no easy way to teardown the integration (all media files and posts using media files will be invalid). So this is always a one time thing.

tests/unit/test_charm_2.py
weiiwang01: This is PHP array serialization, currently I am copying all related methods from the original branch. But I have already developed a better way to do this `wp option update <json> --format=json`. This better method is only used in openstack swift plugin, but I will use this to replace all old methods like this line you mentioned.

tests/integration/test_core.py
```diff
@@ -255,3 +265,103 @@ async def test_wordpress_plugin_installation_error(
...
+        image = image_buf.getvalue()
+        wp = WordpressClient(
+            host=unit_ip, username="admin", password=default_admin_password, is_admin=True
+        )
+        image_urls = wp.upload_media(filename=filename, content=image)
+        swift_object_list = [
+            o["name"] for o in swift_conn.get_container(container, full_listing=True)[1]
+        ]
+        assert (
+            any(nonce in f for f in swift_object_list)
```
mthaddon, line 354: rather than `any(nonce` could/should this be `any(filename`? Seems like it's a bit more specific and clearer that we're checking for the file existing (if I'm understanding correctly)

tests/integration/test_core.py
```diff
@@ -255,3 +265,103 @@ async def test_wordpress_plugin_installation_error(
...
+        ops_test: pytest_operator.plugin.OpsTest,
+        application_name,
+        default_admin_password,
+        unit_ip_list,
+        openstack_environment
+):
+    """
+    arrange: after WordPress charm has been deployed, db relation established and openstack swift
+        server ready
+    act: update charm configuration for openstack object storage plugin
```
mthaddon, line 281: I feel like either here or in the assert we should be saying we're going to upload an image to swift from each unit and confirm it's downloadable as well.

tests/integration/wordpress_client.py
```diff
@@ -184,3 +186,32 @@ def list_plugins(self):
             except_status_code=200,
         )
         return [p["plugin"].split("/")[0] for p in response.json()]
+
+    def upload_media(
+            self, filename: str, content: bytes, mimetype: str = None
+    ) -> typing.List[str]:
+        """Upload a media file (image/video)
+
+        Return url of different sized for the uploaded file on WordPress.
```
mthaddon, line 195: How about "Return URL of resized media file after upload to WordPress."? ("url" should be capitalised here, and "different sized" is a little unclear - feel free to reword as you see fit).

tests/unit/test_charm_2.py
```diff
@@ -126,14 +173,26 @@ def mock_test_database_connectivity(_self):
...
             _current_wp_config=mock_current_wp_config,
             _push_wp_config=mock_push_wp_config,
             _remove_wp_config=mock_remove_wp_config,
             _run_wp_cli=mock_run_wp_cli,
             _test_database_connectivity=mock_test_database_connectivity,
-            _DB_CHECK_INTERVAL=0
+            _DB_CHECK_INTERVAL=0,
+            _apache_config_is_enabled=mock_apache_config_is_enabled,
+            _apache_enable_config=mock_apache_enable_config,
+            _apache_disable_config=mock_apache_disable_config
```
mthaddon, line 195: Can you add a trailing "," here so future additions won't require a line change here?

tests/unit/test_charm_2.py
```diff
@@ -630,3 +698,101 @@ def test_plugin_reconciliation(self):
...
+
+        self.assertEqual(
+            self.patch.get_active_plugins(db_host="config_db_host", db_name="config_db_name"),
+            {plugin},
+            f"{plugin} should be activated after {plugin_config_keys} being set"
+        )
+        self.assertEqual(
+            self.patch.get_options(db_host="config_db_host", db_name="config_db_name"),
+            excepted_options,
+            "akismet options should be set correctly"
```
mthaddon, line 726: Should this be explicitly referencing akismet? I think it should be whichever plugin is being used here, right?

mthaddon (commented):
Some comments inline. Great work so far!

tests/integration/test_core.py
weiiwang01: WordPress may change the filename (e.g. .jpg to .jpeg, add a 2 to the end of filename if filename is duplicate), so here only check the randomly generated nonce to be safe.

tests/integration/test_core.py
weiiwang01: Added, thanks!

tests/integration/wordpress_client.py
weiiwang01: Updated to `Return URL of the original image and resized images for the uploaded file on WordPress.`

WordPress will create a URL for the original image and a series of resized images.

tests/unit/test_charm_2.py
weiiwang01: Fixed, thanks!

tests/unit/test_charm_2.py
weiiwang01: Fixed, thanks!

tests/integration/test_core.py
mthaddon: Ok, thanks

mthaddon (approved):
Thanks for all the changes - looks good from my side

src/charm.py
weiiwang01: Added, thanks.

src/charm.py
```diff
@@ -729,6 +733,12 @@ def _on_leader_elected_replica_data_handler(self, event):
         Only the leader can update the data shared with all replicas. Leader should check if
         the data exist when leadership is established, generate required data and set it in
         the peer relation if not.
+
+        Args:
+            event: required by ops framework, not used.
+
+        Returns:
```
jdkandersson, line 740: The returns is usually optional if there is nothing being returned

jdkandersson (approved):
Thanks for updating the docstrings!

src/charm.py
weiiwang01: I will remove them in the future, thanks!

PR #10, opened 2022-10-17
Add integration tests for Akismet and OpenID plugins

weiiwang01:


src/charm.py
```diff
@@ -52,9 +52,9 @@ class _ReplicaRelationNotReady(Exception):
         'mscom',
         'thematic',
         'twentyeleven',
-        'twentynineteen',
         'twentytwenty',
         'twentytwentyone',
+        'twentytwentytwo',
```
arturo-seijas, line 57: Why this change?

weiiwang01: I bumped the WordPress version to 6.0.2, the default theme has changed so I need to update that in charm.

arturo-seijas (approved):
LGTM

tests/integration/conftest.py
```diff
@@ -102,3 +102,27 @@ def openstack_environment(request):
     openstack_conf = configparser.ConfigParser()
     openstack_conf.read_string("[DEFAULT]\n" + rc_file)
     return {k.upper(): v for k, v in openstack_conf["DEFAULT"].items()}
+
+
+@pytest.fixture
+def akismet_api_key(request):
+    """The Akismet API key, in str"""
+    api_key = request.config.getoption("--akismet-api-key")
+    assert api_key, "Akismet API key should not be empty"
```
jdkandersson, line 111: Do you think the assert message should include a hint, e.g., with the parameter name and expected value? E.g., `Akismet API key should not be empty, please include it in the --akismet-api-key parameter`

tests/integration/conftest.py
```diff
@@ -102,3 +102,27 @@ def openstack_environment(request):
...
+    return api_key
+
+
+@pytest.fixture
+def openid_username(request):
+    """The OpenID username for testing the OpenID plugin"""
+    openid_username = request.config.getoption("--openid-username")
+    assert openid_username, "OpenID username should not be empty"
```
jdkandersson, line 119: Do you think the assert message should include a hint, e.g., with the parameter name and expected value? E.g., `OpenID username should not be empty, please include it in the --openid-username parameter`

tests/integration/conftest.py
```diff
@@ -102,3 +102,27 @@ def openstack_environment(request):
...
+    return openid_username
+
+
+@pytest.fixture
+def openid_password(request):
+    """The OpenID username for testing the OpenID plugin"""
+    openid_password = request.config.getoption("--openid-password")
+    assert openid_password, "OpenID password should not be empty"
```
jdkandersson, line 127: Do you think the assert message should include a hint, e.g., with the parameter name and expected value? E.g., `OpenID password should not be empty, please include it in the --openid-password parameter`

tests/integration/test_core.py
```diff
@@ -353,3 +373,69 @@ async def test_openstack_object_storage_plugin(
...
+    await application.set_config({"wp_plugin_akismet_key": akismet_api_key})
+    await ops_test.model.wait_for_idle()
+
+    for unit_ip in unit_ip_list:
+        wp = WordpressClient(
+            host=unit_ip, username="admin", password=default_admin_password, is_admin=True
+        )
+        post = wp.create_post(secrets.token_hex(8), secrets.token_hex(8))
+        wp.create_comment(
+            post_id=post["id"], post_link=post["link"], content="akismet-guaranteed-spam"
```
jdkandersson, line 401: Nice, it is good that Akismet seems to have an inbuilt way of testing it

tests/integration/test_core.py
```diff
@@ -353,3 +373,69 @@ async def test_openstack_object_storage_plugin(
...
+    unit_ip_list,
+    akismet_api_key,
+):
+    """
+    arrange: after WordPress charm has been deployed, db relation established
+    act: update charm configuration for Akismet plugin
+    assert: Akismet plugin should be activated and spam detection function should be working
+    """
+    application = ops_test.model.applications[application_name]
+    await application.set_config({"wp_plugin_akismet_key": akismet_api_key})
```
jdkandersson, line 392: Do you think the configuration should be reset after the test completes so as not to impact other integration tests? If so, do you think this should be in a fixture to guarantee that the configuration change is reverted?

tests/integration/test_core.py
```diff
@@ -353,3 +373,69 @@ async def test_openstack_object_storage_plugin(
...
+    openid_username,
+    openid_password,
+):
+    """
+    arrange: after WordPress charm has been deployed, db relation established
+    act: update charm configuration for Akismet plugin
+    assert: An Ubuntu One OpenID account should be able to associate with a WordPress user
+    """
+    application = ops_test.model.applications[application_name]
+    await application.set_config({"wp_plugin_openid_team_map": "site-sysadmins=administrator"})
```
jdkandersson, line 427: Similar to the above, do you think this configuration should be reverted after the test completes? If so, do you think it should be done in a fixture to guarantee that it is reverted?

tests/integration/wordpress_client.py
```diff
@@ -108,7 +108,7 @@ def _gen_wp_rest_nonce(self):
         nonce = json.loads(re.findall('var wpApiSettings = ([^;]+);', new_post_page)[0])["nonce"]
         return nonce
 
-    def _create_post(self, title: str, content: str):
+    def create_post(self, title: str, content: str):
         """Create a WordPress post using wp-json API, return post object"""
```
jdkandersson, line 112: Do you think it is worth updating the docstring to our new standard and to define a return type?

tests/integration/wordpress_client.py
```diff
@@ -118,7 +118,7 @@ def _create_post(self, title: str, content: str):
         )
         return response.json()
 
-    def _create_comment(self, post_id: int, post_link: str, content: str):
+    def create_comment(self, post_id: int, post_link: str, content: str):
         """Add a comment to a WordPress post using HTML form, return url link of the new comment"""
```
jdkandersson, line 122: Do you think it is worth updating the docstring to our new standard and to define a return type?

tests/integration/wordpress_client.py
```diff
@@ -166,6 +166,14 @@ def list_plugins(self):
         )
         return [p["plugin"].split("/")[0] for p in response.json()]
 
+    def list_comments(self, status='approve', post_id: int = None):
+        """List all comments in the WordPress site"""
```
jdkandersson, line 170: We have a new standard for docstrings, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#docstrings

tests/integration/wordpress_client.py
```diff
@@ -192,3 +200,72 @@ def upload_media(self, filename: str, content: bytes, mimetype: str = None) -> t
         if media["source_url"] not in image_urls:
             image_urls.append(media["source_url"])
         return image_urls
+
+    def associate_ubuntu_one(self, username, password):
+        openid_setting_url = f"http://{self.host}/wp-admin/users.php?page=your_openids"
```
jdkandersson, line 205: We have a new standard for docstrings, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#docstrings

tests/integration/wordpress_client.py
```diff
@@ -166,6 +166,14 @@ def list_plugins(self):
         )
         return [p["plugin"].split("/")[0] for p in response.json()]
 
+    def list_comments(self, status='approve', post_id: int = None):
```
jdkandersson, line 169: We have a new standard for type hints, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#type-hints

tests/integration/wordpress_client.py
```diff
@@ -192,3 +200,72 @@ def upload_media(self, filename: str, content: bytes, mimetype: str = None) -> t
         if media["source_url"] not in image_urls:
             image_urls.append(media["source_url"])
         return image_urls
+
+    def associate_ubuntu_one(self, username, password):
```
jdkandersson, line 204: We have a new standard for type hints, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#type-hints

tests/integration/wordpress_client.py
```diff
@@ -192,3 +200,72 @@ def upload_media(self, filename: str, content: bytes, mimetype: str = None) -> t
...
+            data=openid_args,
+        )
+        csrf_token = re.findall(
+            "<input type='hidden' name='csrfmiddlewaretoken' value='([^']+)' />", login_page.text
+        )[0]
+        login_link = re.findall(
+            '<a id="login-link" data-qa-id="login_link" href="([^"]+)" class="p-link--soft">',
+            login_page.text,
+        )[0]
+        login_url = "https://login.ubuntu.com" + login_link
```
jdkandersson, line 235: f-string?

tests/integration/wordpress_client.py
```diff
@@ -192,3 +200,72 @@ def upload_media(self, filename: str, content: bytes, mimetype: str = None) -> t
...
+                'fullname': 'on',
+                'ok': '',
+                'yes': '',
+                'openid.usernamesecret': '',
+            },
+            headers={"Referer": confirm_page.url},
+            except_status_code=200,
+        )
+
+    def list_associated_ubuntu_one_accounts(self):
```
jdkandersson, line 266: We have a new standard for docstrings, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#docstrings

We have a new standard for type hints, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#type-hints

jdkandersson (changes requested):

tests/integration/conftest.py
weiiwang01: Fixed, thanks!

tests/integration/conftest.py
weiiwang01: Fixed, thanks!

tests/integration/conftest.py
weiiwang01: Fixed, thanks!

tests/integration/wordpress_client.py
weiiwang01: Fixed, thanks!

tests/integration/wordpress_client.py
weiiwang01: Fixed, thanks!

tests/integration/wordpress_client.py
weiiwang01: Fixed, thanks!

tests/integration/wordpress_client.py
weiiwang01: Fixed, thanks!

tests/integration/wordpress_client.py
weiiwang01: Fixed, thanks!

tests/integration/test_core.py
weiiwang01: There's a fixture for config and revert config using fixture parameter. But this test case requires `akismet_api_key` which is a fixture, and you can't pass a fixture to a fixture parameter.

PR #11, opened 2022-10-20
Add integration tests for ingress and some cleanup

weiiwang01:


src/charm.py
```diff
@@ -165,6 +155,14 @@ def ingress_config(self):
             ingress_config["tls-secret-name"] = tls_secret_name
```
arturo-seijas, line 155: Should we set the secret here? This can be directly configured directly the ingress without having wordpress sitting in the middle, providing a more standard way across deployments (and easier to automate in the future)

arturo-seijas (commented):
I've left some comments. Also, I find this PR a little bigger than ideal. May I suggest you split it in more PRs in the future for easier review?
Nice work, btw

src/charm.py
```diff
@@ -146,15 +136,15 @@ def __init__(self, *args, **kwargs):
...
         self.framework.observe(self.on.wordpress_pebble_ready, self._reconciliation)
         self.framework.observe(self.on["wordpress-replica"].relation_changed, self._reconciliation)
         self.framework.observe(self.db.on.database_changed, self._reconciliation)
 
     @property
     def ingress_config(self):
-        blog_hostname = self.state.blog_hostname
+        blog_hostname = self.model.config["blog_hostname"] or self.app.name
         ingress_config = {
             "service-hostname": blog_hostname,
```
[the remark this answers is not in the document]
weiiwang01: The port is not adjustable on the original sidecar branch, so it's the only possible port right now.

src/charm.py
weiiwang01: The original version of the charm has a config called `tls_secret_name`, this is solely for the backward compatibility's sake. Maybe I will remove this in the future.

weiiwang01:
> I've left some comments. Also, I find this PR a little bigger than ideal. May I suggest you split it in more PRs in the future for easier review?
> Nice work, btw

Thanks! I will try to keep the PR size down in the future.

arturo-seijas (approved):
LGTM

PR #12, opened 2022-10-25
Fix the Launchpad OpenID integration

weiiwang01:
WordPress launchpad integration (login using launchpad account, team-role mapping) requires three plugins to work: `openid`, `wordpress-launchpad-integration` and `wordpress-teams-integration`.

Setting the `wp_plugin_openid_team_map` charm config in the current stable version will only activate one plugin `openid` and update one of the WordPress options for  `wordpress-teams-integration`. That means after setting `wp_plugin_openid_team_map`, some undocumented manual operation is still required.

This PR consolidate all manual operation into the charm, so after setting the `wp_plugin_openid_team_map` charm config, WordPress launchpad integration will be set up completed.

Pining Ubuntu version to 20.04, WordPress version to 5.9, and openid plugin version to 3.5 due to the compatibility issue of `wordpress-launchpad-integration` and `wordpress-teams-integration`.


src/charm.py
```diff
@@ -946,58 +947,86 @@ def _plugin_akismet_reconciliation(self):
...
+                raise exceptions.WordPressBlockedStatusException(
+                    f"Unable to config openid plugin, {result.message}"
+                )
+
         if not openid_team_map:
+            result = self._wp_option_update("users_can_register", "0")
+            check_result()
             result = self._deactivate_plugin(
-                "openid", ["openid_required_for_registration", "openid_teams_trust_list"]
+                'wordpress-teams-integration', ["openid_teams_trust_list"]
```
arturo-seijas, line 1003: Does this pass linting? I'd expect double quotes here

tox.ini
```diff
@@ -43,11 +44,11 @@ setenv =
 
 
 [testenv:black]
-commands = black --skip-string-normalization --line-length=100 src/ tests/
+commands = black --skip-string-normalization --line-length=99 src/ tests/
 deps = black
 
 [testenv:lint]
-commands = flake8 src/ tests/
+commands = flake8 --ignore=W503,C901 src/ tests/
```
arturo-seijas, line 51: Would it be possible to ignore this for specific lines of code as in https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors?

src/charm.py
weiiwang01: It was causing by the current black configuration. Fixed, thanks!

tox.ini
weiiwang01: I use the inline ignore to ignore C901: the function complexity warning. In the case of W503, due to a bug in flake, W503 conflicts with W504, so one of the two must be disabled completely.

tox.ini
```diff
@@ -43,11 +44,11 @@ setenv =
 
 
 [testenv:black]
-commands = black --skip-string-normalization --line-length=100 src/ tests/
+commands = black src/ tests/
 deps = black
 
 [testenv:lint]
-commands = flake8 src/ tests/
+commands = flake8 --ignore=W503 src/ tests/
```
arturo-seijas, line 51: Would it be possible to ignore this for specific lines of code as in https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors or maybe fix it?

tox.ini
arturo-seijas: Understood. Thnx

arturo-seijas (approved):
LGTM

arturo-seijas (approved):
LGTM

PR #14, opened 2022-10-28
Add continuous integration GitHub workflows

weiiwang01:
Demo: https://github.com/weiiwang01/wordpress-k8s-operator/pull/18


.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Run integration tests
+        run: > 
+          tox -e integration -- 
+          --openstack-rc ${GITHUB_WORKSPACE}/openrc
+          --akismet-api-key ${{ secrets.TEST_AKISMET_API_KEY }}
+          --openid-username ${{ secrets.TEST_OPENID_USERNAME }}
+          --openid-password ${{ secrets.TEST_OPENID_PASSWORD }}
+          --kube-config ${GITHUB_WORKSPACE}/kube-config
+          --launchpad-team ${{ secrets.TEST_LAUNCHPAD_TEAM }}
```
arturo-seijas, line 52: You're missing some blank lines here

.github/workflows/test.yaml
```diff
@@ -0,0 +1,9 @@
+name: Tests
+
+on:
+  pull_request:
+
+jobs:
+  unit-tests:
+    uses: canonical/operator-workflows/.github/workflows/test.yaml@v1.0.0
+    secrets: inherit
```
arturo-seijas, line 9: You're missing some blank lines here

jdkandersson (changes requested):
Looking good, thanks for making these changes!

src/exceptions.py
```diff
@@ -13,6 +13,8 @@
 ]
 
 
+# This exception is used to signal the early termination of a reconciliation process,
+# thus not an error (N818), same for all the subclasses.
```
jdkandersson, line 17: Thanks for adding the note, I don't yet follow why it isn't an error? If the reconciliation process terminates early, is that not an unexpected behaviour and thus an error?

.github/workflows/integration_test.yaml
weiiwang01: Fixed, thanks!

.github/workflows/test.yaml
weiiwang01: Fixed, thanks!

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Create OpenStack credential file
+        run: echo "${{ steps.setup-devstack-swift.outputs.credentials }}" > openrc
+
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          provider: microk8s
+
+      - name: Enable microk8s plugins
+        run: sudo microk8s enable registry hostpath-storage ingress
```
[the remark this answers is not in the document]
weiiwang01: Fixed, thanks!

pyproject.toml
```diff
@@ -1,2 +1,39 @@
...
+exclude_dirs = ["/venv/"]
+[tool.bandit.assert_used]
+skips = ["*/*test.py", "*/test_*.py"]
+
+# Testing tools configuration
+[tool.coverage.run]
+branch = true
+
+[tool.coverage.report]
+fail_under = 80
```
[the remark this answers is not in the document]
weiiwang01: Yes, this is a problem that need to be addressed, but I think this is out of the scope of this PR, so I will create a new PR  to adding more unit tests.

src/exceptions.py
```diff
@@ -1,3 +1,7 @@
+# Copyright 2022 Canonical Ltd.
+# Licensed under the GPLv3, see LICENCE file for details.
+
+"""User-defined exceptions used by WordPress charm."""
 import ops.model
```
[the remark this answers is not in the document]
weiiwang01: I add the empty line anyway, thanks!

src/exceptions.py
weiiwang01: In some cases, the reconciliation can terminates early like configuration hasn't been updated or the relation is not ready yet. The charm will go into the waiting state but I don't thing this count as an error because I think they are excepted in the early stage of the charm lifecycle. That's why I use exception instead of error here.

tests/integration/requirements.txt
```diff
@@ -1,4 +1,4 @@
-juju
+juju==3.0.2
```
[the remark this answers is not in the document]
weiiwang01: This is because of the bug in v3.0.3 of libjuju, that's fixed now. I just unpinned the dependency.
Thanks!

src/exceptions.py
jdkandersson: Ah I see, do you think it is worth including that information in the comment? That extra information cleared it up for me

jdkandersson (approved):

src/exceptions.py
weiiwang01: Done

jdkandersson (approved):

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  pull_request:
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup Devstack Swift
```
arturo-seijas, line 12: Everything except this step can be run with the current pipeline. I'll try to find a way to inject steps but I don't think it'll be straightforward.
I'll keep you posted

arturo-seijas (approved):
LGTM

PR #17, opened 2022-11-14
Convert unit test framework from unittest to pytest

weiiwang01:


merkata (approved):
Nice work!

pyproject.toml
```diff
@@ -8,7 +8,7 @@ skips = ["*/*test.py", "*/test_*.py"]
 branch = true
 
 [tool.coverage.report]
-fail_under = 80
+fail_under = 90
```
jdkandersson, line 11: 🥳

tests/unit/conftest.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@pytest.fixture(scope="function")
+def app_name():
+    """The name of the charm application."""
+    return "wordpress-k8s"
+
+
+@pytest.fixture(scope="function")
+def setup_replica_consensus(harness, app_name):
```
jdkandersson, line 44: For this fixture and others, it can be helpful to type these (e.g., `def setup_replica_consensus(harness: ops.testing.Harness, app_name: str):`), this will provide auto-complete in some editors and get mypy to check for potential bugs. In fixtures and testing, usually return values are pretty pointless since pytest does magic with the return values so I usually omit them

tests/unit/test_charm.py
```diff
@@ -7,651 +7,488 @@
...
-        self.harness.set_leader(True)
-        self.assertDictEqual(
-            consensus,
-            self.harness.get_relation_data(replica_relation_id, self.app_name),
-            "consensus once established should not change after leadership changed",
-        )
-
-    @staticmethod
-    def _example_db_info():
+def test_generate_wp_secret_keys(harness):
```
jdkandersson, line 16: For this and other test functions, it can be helpful to type the arguments (e.g., `harness: ops.testing.Harness`) to help with auto-complete in some editors and to help mypy find potential bugs

tests/unit/test_charm.py
```diff
@@ -7,651 +7,488 @@
...
-        return event_mock
-
-    def test_get_initial_password_action_before_replica_consensus(self):
-        """
-        arrange: before peer relation established but after charm created
-        act: run get-initial-password action
-        assert: get-initial-password action should fail
+    harness.begin_with_initial_hooks()
+
+    assert {None} == set(
```
jdkandersson, line 97: This is a picky comment and this could have already been coded in this way (there could also be other instances of this in the code which might be worth checking), it can be helpful to consistently put the value being checked on the left and the expected value on the right of the equality in `assert`, in this case it is obvious, in others it might not always be clear which value the test is asserting to be right if the assert fails

tests/unit/conftest.py
weiiwang01: Added, thanks!

tests/unit/test_charm.py
weiiwang01: Added, thanks!

tests/unit/test_charm.py
weiiwang01: Fixed, thanks!

tests/unit/conftest.py
```diff
@@ -72,15 +72,15 @@ def example_db_info():
...
+def setup_db_relation(harness: ops.testing.Harness, example_db_info: dict):
     """Yields a function that can be used to set up db relation.
 
     After calling the yielded function, a db relation will be set up. If ``db_info`` not provided
     during the invocation, example_db_info will be used as the relation data. Return a tuple of
     relation id and the relation data.
     """
 
-    def _setup_db_relation(db_info=None):
+    def _setup_db_relation(db_info: dict = None):
```
jdkandersson, line 83: Should this be `dict | None` or `typing.Union[dict, None]` for before Python 3.10?

weiiwang01: I removed this function argument after I found out that it's used anywhere. Thanks!

jdkandersson (approved):

PR #18, opened 2022-11-14
Enable OWASP ModSecurity Core Rule Set (CRS)

amandahla:
Add new configuration "use_nginx_ingress_modsec" (default true) to enable Nginx Ingress ModSecurity.

Add 1 unit test (only checks config change) and 1 integration test (reference [here](https://docs.nginx.com/nginx-waf/admin-guide/nginx-plus-modsecurity-waf-owasp-crs/#disabling-blocking-of-requests-based-on-the-useragent-header))

Update Ingress lib.
```charmcraft fetch-lib charms.nginx_ingress_integrator.v0.ingress```

WIP investigation: When changing the config to False, the ModSecurity isn't disabled.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -132,6 +132,8 @@ class WordpressCharm(CharmBase):
             relation_db_password=None,
         )
 
+        logger.info(self.ingress_config)
+
         self.ingress = IngressRequires(self, self.ingress_config)
 
         self.framework.observe(
@@ -160,6 +162,13 @@ class WordpressCharm(CharmBase):
         tls_secret_name = self.model.config["tls_secret_name"]
         if tls_secret_name:
             ingress_config["tls-secret-name"] = tls_secret_name
+
+        owasp_modsecurity_crs = False
+        data = self.config.get("use_nginx_ingress_modsec", None)
+        if data is not None and data != "":
+            owasp_modsecurity_crs = data
+        ingress_config["owasp-modsecurity-crs"] = owasp_modsecurity_crs
+
         return ingress_config
 
     def _update_ingress_config(self, _event):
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by jdkandersson** [comment]
@amandahla usually we create branches rather than using forks for our development since you should have maintainer access

**Comment by weiiwang01** [comment]
> @amandahla usually we create branches rather than using forks for our development since you should have maintainer access

Hi, I think what David said is that, there are some GitHub secrets that's used in the integration tests in this repository that's not accessible from a forked repo. So all PRs from forked repos will fail, and that's why the integration test failed here.

We will fix this problem in the future. You can create a new branch inside this repo and start a new PR to bypass this problem for now.

File: src/charm.py [hunk 1]
```diff
@@ -160,6 +160,8 @@ def ingress_config(self):
         tls_secret_name = self.model.config["tls_secret_name"]
         if tls_secret_name:
             ingress_config["tls-secret-name"] = tls_secret_name
+        ingress_config["owasp-modsecurity-crs"] = self.model.config["use_nginx_ingress_modsec"]
```
mthaddon, line 163: I think we need to be a little careful here - I'm not sure we can turn this on without some custom exclusions. See https://code.launchpad.net/~sajoupa/charm-k8s-wordpress/+git/charm-k8s-wordpress/+merge/414554 for more information. This was preventing post requests from succeeding per Ubuntu RT#37351: "Ubuntu Fridge - unable to post (UWN 719)"

weiiwang01:
Supersede by [canonical/wordpress-k8s-operator#26](https://github.com/canonical/wordpress-k8s-operator/pull/26).

PR #19, opened 2022-11-15
Add WordPress health check

weiiwang01:


mthaddon:
Can we take a look at the failing integration test before merging? I don't think it's related to this change, but it'd be good to confirm what it is.

weiiwang01:
> Can we take a look at the failing integration test before merging? I don't think it's related to this change, but it'd be good to confirm what it is.

It seems like one of the integration tests is using `wait_for_idle` but without the `status` parameter. So when the charm was performance database initialization in the maintenance status, it took more than 15 seconds, the default idle interval, and causes the wait to exit early.

Usually, database initialization only takes a few seconds, that's why this never happens before. But it is taking more time this time, probably due to the database performance issue.

jdkandersson (approved):

PR #21, opened 2022-11-17
Rotate secrets

yhaliaw:
Add action for rotating secrets.
Enable pebble connection unit test. See https://juju.is/docs/sdk/testing#heading--simulate-can-connect.


yhaliaw:
For the Inclusive naming check, the failures are from the included library nginx-ingress-integrator.
It may be best, for it to be fixed in that repo, instead of change the lib files in this repo.

pyproject.toml
```diff
@@ -8,7 +8,7 @@ skips = ["*/*test.py", "*/test_*.py"]
 branch = true
 
 [tool.coverage.report]
-fail_under = 90
+fail_under = 85
```
yhaliaw, line 11: Temporary decrease. This pull request should have 100% code coverage for the code added.
In the future, more test will be added to cover prior code.

gregory-schiano: Should be discussed with the team, but I propose to never touch the coverage target to ensure the PR cannot be merged in the main branch with this change

src/charm.py
```diff
@@ -179,6 +182,29 @@ def _on_get_initial_password_action(self, event):
             logger.error("Action get-initial-password failed. Replica consensus not exists")
             event.fail("Default admin password has not been generated yet.")
 
+    def _on_rotate_wordpress_secrets_action(self, event):
+        """Handle the rotate-wordpress-secrets action."""
+        if self._replica_consensus_reached():
+            if self.unit.is_leader():
+                # Update the secrets in peer relation.
+                replica_relation_data = self._replica_relation_data()
+                secrets = self._generate_wp_secret_keys()
```
gregory-schiano, line 191: It looks like this method will also rotate the password salts which as a consequence will lead to all existing passwords behind unusable for users.
Is it really expected ? It could totally break the wordpress

tests/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2022 Canonical Ltd.
+# Licensed under the GPLv3, see LICENCE file for details.
```
arturo-seijas, line 2: This license headers is not the one we should be using. Use
```
# Copyright 2022 Canonical Ltd.
# See LICENSE file for licensing details.
```
instead.
I assume you got it from another file. Can you please change all occurrences?

src/charm.py
yhaliaw: I confirmed with @weiiwang01 that this is the intended behavior.
This is the same as the original PR on launchpad this PR is based on.
https://code.launchpad.net/~jsimpso/charm-k8s-wordpress/+git/charm-k8s-wordpress/+merge/419738

tests/__init__.py
yhaliaw: I have changed all occurrences.
I have also notice the python file `ingress.py` under `lib/charms/nginx_ingress_integrator/v0` does not start with any licensing info. Is this fine?

pyproject.toml
yhaliaw: Done.

src/charm.py
weiiwang01: I think these are just the salt for cookies, so updating these will not reset user's password.

Here's some explanation from WordPress:
```php

/**#@+
 * Authentication unique keys and salts.
 *
 * Change these to different unique phrases! You can generate these using
 * the {@link https://api.wordpress.org/secret-key/1.1/salt/ WordPress.org secret-key service}.
 *
 * You can change these at any point in time to invalidate all existing cookies.
 * This will force all users to have to log in again.
 *
 * @since 2.6.0
 */
define( 'AUTH_KEY',         'put your unique phrase here' );
define( 'SECURE_AUTH_KEY',  'put your unique phrase here' );
define( 'LOGGED_IN_KEY',    'put your unique phrase here' );
define( 'NONCE_KEY',        'put your unique phrase here' );
define( 'AUTH_SALT',        'put your unique phrase here' );
define( 'SECURE_AUTH_SALT', 'put your unique phrase here' );
define( 'LOGGED_IN_SALT',   'put your unique phrase here' );
define( 'NONCE_SALT',       'put your unique phrase here' );

```

tests/__init__.py
arturo-seijas: It might be. Since those libraries are downloaded from charmhub we'd have to update the appropriate sources.
Thank you

mthaddon: https://github.com/canonical/nginx-ingress-integrator-operator/pull/37 to fix the lack of copyright notice in that library file.

mthaddon:
> For the Inclusive naming check, the failures are from the included library nginx-ingress-integrator. It may be best, for it to be fixed in that repo, instead of change the lib files in this repo.

The nginx-ingress-integrator charm uses "whitelist" because that's what upstream nginx-ingress uses. It's fine to just add this file to an exclusion list in `.woke.yaml`.

arturo-seijas:
> For the Inclusive naming check, the failures are from the included library nginx-ingress-integrator. It may be best, for it to be fixed in that repo, instead of change the lib files in this repo.

You can see how to ignore the file [here](https://github.com/canonical/discourse-k8s-operator/blob/main/.woke.yaml)

weiiwang01:
> > For the Inclusive naming check, the failures are from the included library nginx-ingress-integrator. It may be best, for it to be fixed in that repo, instead of change the lib files in this repo.
> 
> You can see how to ignore the file [here](https://github.com/canonical/discourse-k8s-operator/blob/main/.woke.yaml)

I added a `.woke.yaml` file and fixed the problem causing the integration test to fail in this PR https://github.com/canonical/wordpress-k8s-operator/pull/23. You can rebase after that PR is merged.

src/charm.py
```diff
@@ -179,6 +182,29 @@ def _on_get_initial_password_action(self, event):
             logger.error("Action get-initial-password failed. Replica consensus not exists")
             event.fail("Default admin password has not been generated yet.")
 
+    def _on_rotate_wordpress_secrets_action(self, event):
```
jdkandersson, line 185: We have a standard for including type hints for functions, do you think it is applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#type-hints

src/charm.py
```diff
@@ -179,6 +182,29 @@ def _on_get_initial_password_action(self, event):
...
+        """Handle the rotate-wordpress-secrets action."""
+        if self._replica_consensus_reached():
```
jdkandersson, line 187: There is a way to refactor this to reduce nesting and improve readability, pseudo code below. It reduces the context that the reader has to maintain in their mind as they read the code. Do you think it would improve the code?

```Python
if not self._replica_consensus_reached():
    <log and fail the action>

if not self.unit.is_leader():
    <log and fail the action>

<code for action execution>
```

src/charm.pyi
```diff
@@ -1,5 +1,5 @@
 # Copyright 2022 Canonical Ltd.
```
jdkandersson, line 1: Do the type hints have to be updated with the addition of a new function?

tests/unit/test_charm.py
```diff
@@ -342,6 +345,87 @@ def test_get_initial_password_action(
...
+
+    harness.charm._on_rotate_wordpress_secrets_action(action_event_mock)
+
+    # Technically possible to generate the same passwords, but extremely unlikely.
+    assert (
+        old_relation_data
+        != harness.model.get_relation("wordpress-replica").data[harness.charm.app]
+    ), "password are same from before rotate"
+
+    assert action_event_mock.set_results.mock_calls == [unittest.mock.call({"result": "ok"})]
```
jdkandersson, line 425: I believe mocks have functions such as `assert_called_once_with` which can be easier to read: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with

jdkandersson (commented):
Thank you for creating the PR, do you think an integration test would be worth it?

src/charm.pyi
yhaliaw: I have updated type hints.

src/charm.py
yhaliaw: Thanks for this suggestion.
Updated.

tests/unit/test_charm.py
jdkandersson: The updated functions look good, could this line be changed to: `assert action_event_mock.set_results.assert_called_once_with({"result": "ok"})`?

yhaliaw: I have updated all existing unit test with `Mock` with `assert_called_once_with` and `assert_not_called`.

src/charm.py
```diff
@@ -179,6 +182,29 @@ def _on_get_initial_password_action(self, event):
             logger.error("Action get-initial-password failed. Replica consensus not exists")
             event.fail("Default admin password has not been generated yet.")
 
+    def _on_rotate_wordpress_secrets_action(self, event):
+        """Handle the rotate-wordpress-secrets action."""
```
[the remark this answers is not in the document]
yhaliaw: Updated the docstring.

yhaliaw:
> Thank you for creating the PR, do you think an integration test would be worth it?

I think it would be worth it. 
I think having integration with leader and follows, i.e. more than one unit, would be a good idea.

I will look into it once the integration test fix is merged into master, and this branch is rebased on that fix.

jdkandersson:
> > Thank you for creating the PR, do you think an integration test would be worth it?
> 
> I think it would be worth it. I think having integration with leader and follows, i.e. more than one unit, would be a good idea.
> 
> I will look into it once the integration test fix is merged into master, and this branch is rebased on that fix.

Ok, if we have a dependency we can skip that for now and maybe just create a follow up story so we pick that up in the future?

jdkandersson (approved):

jdkandersson (dismissed):

arturo-seijas (approved):
LGTM

weiiwang01 (approved):

PR #23, opened 2022-11-17
Fix the health check problem during the service start-up

weiiwang01:
The current charm version start-up sequence is:

1. add the layer to pebble
2. initialize the WordPress database
3. start the web server

So when the readiness check defined in the pebble layer was added, the server is not started immediately. Sometimes the in-between steps will take more than 30 seconds, which will cause the health check to fail and Kubernetes will restart the container, causing a series of trouble.

Change the start-up sequence to avoid this problem:

1. initialize the WordPress database
2. add the layer to pebble
3. start the web server

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -680,9 +680,10 @@ class WordpressCharm(CharmBase):
     def _check_addon_type(self, addon_type):
         """Check if addon_type is one of the accepted addon types (theme/plugin).
 
-        Raise a AssertError if not.
+        Raise a ValueError if not.
         """
-        assert addon_type in ("theme", "plugin")
+        if addon_type not in ("theme", "plugin"):
+            raise ValueError(f"Addon type unknown {repr(addon_type)}, accept: (theme, plugin)")
 
     def _wp_addon_list(self, addon_type):
         """List all installed WordPress addons.
@@ -855,7 +856,10 @@ class WordpressCharm(CharmBase):
         Returns:
             An instance of :attr:`charm.WordpressCharm._ExecResult`.
         """
-        assert action in ("activate", "deactivate")
+        if action not in ("activate", "deactivate"):
+            raise ValueError(
+                f"Unknown activation_status {repr(action)}, " "accept (activate, deactivate)"
+            )
         current_plugins = self._wp_addon_list("plugin")
         if not current_plugins.success:
             return self._ExecResult(
--- a/src/charm.py
+++ b/src/charm.py
@@ -586,6 +586,9 @@ class WordpressCharm(CharmBase):
                 if self._wp_is_installed():
                     break
                 else:
+                    self.unit.status = WaitingStatus(
+                        "Waiting for leader unit to initialize database"
+                    )
                     time.sleep(5)
             else:
                 raise exceptions.WordPressBlockedStatusException(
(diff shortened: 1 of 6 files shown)


arturo-seijas (dismissed):
LGTM

src/charm.py
```diff
@@ -301,9 +301,7 @@ def _gen_wp_config(self):
 
         replica_relation_data = self._replica_relation_data()
         for secret_key in self._wordpress_secret_key_fields():
-            secret_value = replica_relation_data.get(secret_key)
-            if not secret_value:
-                raise ValueError(f"{secret_key} value is empty")
+            secret_value = replica_relation_data[secret_key]
```
jdkandersson, line 304: Is there a guarantee that `secret_value` is empty/ None? Just curios since the check for it was removed

src/charm.py
```diff
@@ -584,11 +581,22 @@ def _start_server(self):
 
             if not self._wp_is_installed():
                 self._wp_install()
-            if self._current_wp_config() is None:
-                # For security reasons, never start WordPress server if wp-config.php not exists
-                raise FileNotFoundError(
-                    "required file (wp-config.php) for starting WordPress server not exists"
+        else:
+            for _ in range(60):
```
jdkandersson, line 585: DO you think it would be worth updating the status message of the charm to indicate it is waiting for the database to be ready?

src/charm.py
```diff
@@ -584,11 +581,22 @@ def _start_server(self):
...
+                else:
+                    time.sleep(5)
+            else:
+                raise exceptions.WordPressBlockedStatusException(
+                    "leader unit failed to initialize WordPress database in given time."
                 )
+        if self._current_wp_config() is None:
+            # For security reasons, never start WordPress server if wp-config.php not exists
+            raise FileNotFoundError(
+                "required file (wp-config.php) for starting WordPress server not exists"
```
jdkandersson, line 597: `"required file (wp-config.php) for starting WordPress server does not exists"`?

tests/unit/test_charm.py
```diff
@@ -558,3 +558,64 @@ def additional_check_after_install():
...
+def test_defensive_programing(
+    harness: ops.testing.Harness,
+):
+    """
+    arrange: no arrange.
+    act: invoke some method with incorrect arguments.
+    assert: ValueError should be raised to prevent further execution.
+    """
+    harness.begin()
+    with pytest.raises(ValueError):
```
jdkandersson, line 614: Could this be done using `pytest.mark.parametrize`? E.g., take the function name and argument as a parameter? That way you know all of the cases that fail, currently you will only know the first one that fails since pytest would stop execution after that

jdkandersson (commented):
Is it worth including unit tests and/ or integration tests to check for the problem with health checks we encountered?

src/charm.py
weiiwang01: Instead of using `get`, I am using `__getitem__` which will throw an error when the item does not exist. So the none check here is not needed anymore.

jdkandersson: Makes sense, would it be worth handling that error and giving feedback to the user? I'm guessing the charm would go into error state if `secret_key` is not in `replica_relation_data`?

src/charm.py
weiiwang01: Added, thanks!

src/charm.py
weiiwang01: Fixed, thanks!

tests/unit/test_charm.py
weiiwang01: Fixed, thanks!

src/charm.py
weiiwang01: It shouldn't happen, there must be some unknown internal error if the replica_relation_data doesn't contain the secret keys. So I think it's the best to let it throw as an error.

jdkandersson (approved):

yhaliaw (approved):
LFTM

PR #26, opened 2022-11-30
Add an option to enable modsecurity on ingress

weiiwang01:


lib/charms/nginx_ingress_integrator/v0/ingress.py
```diff
@@ -184,15 +239,31 @@ def _on_relation_changed(self, event):
...
         # configure the ingress.
         self.charm.on.ingress_available.emit()
+
+    def _on_relation_broken(self, event):
+        """Handle a relation-broken event in the ingress relation."""
+        if not self.model.unit.is_leader():
+            return
+
+        # Create an event that our charm can use to remove the ingress resource.
+        self.charm.on.ingress_broken.emit(event.relation)
```
arturo-seijas, line 269: Really picky, but there's a blank line missing here

weiiwang01: Added, thanks!

arturo-seijas (dismissed):
LGTM

gtrkiller (dismissed):
LGTM

tests/integration/test_core.py
```diff
@@ -117,6 +116,113 @@ async def test_mysql_relation(ops_test: pytest_operator.plugin.OpsTest, applicat
...
+    """
+    arrange: after charm deployed, db relation established and openstack swift server ready.
+    act: update charm configuration for openstack object storage plugin.
+    assert: openstack object storage plugin should be installed after the config update and
+        WordPress openstack swift object storage integration should be set up properly.
+        After openstack swift plugin activated, an image file uploaded to one unit through
+        WordPress media uploader should be accessible from all units.
+    """
+    assert ops_test.model
+    swift_conn = swiftclient.Connection(
```
jdkandersson, line 136: Would it be worth including some comments explaining the key steps being taken here?

tests/integration/test_core.py
```diff
@@ -366,114 +473,43 @@ def patched_getaddrinfo(*args):
...
-                    "copy-to-swift": "1",
-                    "serve-from-swift": "1",
-                    "remove-local-file": "0",
-                }
-            )
-        }
-    )
-    await ops_test.model.wait_for_idle()
+    await application.set_config({"use_nginx_ingress_modsec": "true"})
+    await ops_test.model.wait_for_idle(status=ops.model.ActiveStatus.name)  # type: ignore
```
jdkandersson, line 491: Out of curiosity, what is the typing issue here? is it `ops.model.ActiveStatus.name`?

tests/unit/test_charm.py
```diff
@@ -129,6 +129,19 @@ def get_db_info_from_state():
         ), f"database info {db_info_key} should be reset to None after database relation broken"
 
 
+def test_wp_config_before_consensus(harness: ops.testing.Harness):
```
jdkandersson, line 132: Good refactor, this makes it clear this is another test case!

tests/unit/test_charm.py
```diff
@@ -699,16 +707,51 @@ def test_ingress(
...
     assert charm.ingress.config_dict == {
         "service-hostname": app_name,
+        "host": app_name,
+        "service-name": app_name,
+        "name": app_name,
+        "service-port": "80",
+        "port": "80",
+        "tls-secret-name": "tls_secret",
+        "owasp-modsecurity-crs": True,
+        "owasp-modsecurity-custom-rules": 'SecAction "id:900130,phase:1,nolog,pass,t:none,setvar:tx.crs_exclusions_wordpress=1"\n',
```
jdkandersson, line 730: Out of curiosity, how is `id:900130` generated?

weiiwang01: It the rule id to disable certain checks for some well-known applications.
https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/56ff01a8408bd387fe14f7856391c48e245b7937/crs-setup.conf.example#L321-L362

tests/integration/test_core.py
weiiwang01: Yes, `error: Cannot determine type of "name"  [has-type]`

The `name` class attribute is defined as `NotImplemented` in base class and overwritten to string in subclass `ops.model.ActiveStatus`. mypy has some problems handling this case.

tests/integration/test_core.py
weiiwang01: Added, thanks!

jdkandersson: That is excellent, it makes a lot of sense now!

jdkandersson (dismissed):

lib/charms/nginx_ingress_integrator/v0/ingress.py
```diff
@@ -40,11 +48,17 @@
   ingress:
     interface: ingress
 ```
+You _must_ register the IngressRequires class as part of the `__init__` method
+rather than, for instance, a config-changed event handler. This is because
```
gregory-schiano, line 52: You can remove `This is because` to avoid repetition of `because`
`Doing so won't get the current relation changed event, because, ...`

lib/charms/nginx_ingress_integrator/v0/ingress.py
```diff
@@ -40,11 +48,17 @@
...
+doing so won't get the current relation changed event, because it wasn't
+registered to handle the event (because it wasn't created in `__init__` when
```
gregory-schiano, line 54: `since it wasn't created in ...` once more to avoid repetition of `because`

arturo-seijas (dismissed):
LGTM

tests/integration/test_core.py
```diff
@@ -117,6 +116,121 @@ async def test_mysql_relation(ops_test: pytest_operator.plugin.OpsTest, applicat
...
+        swift_conn.head_container(container)
+    except swiftclient.exceptions.ClientException as exc:
+        if exc.http_status == 404:
+            container_exists = False
+        else:
+            raise exc
+    if container_exists:
+        for swift_object in swift_conn.get_container(container, full_listing=True)[1]:
+            swift_conn.delete_object(container, swift_object["name"])
+        swift_conn.delete_container(container)
```
gregory-schiano, line 162: Can't you use the `delete` method from the SwiftService that could do all this in 1 operation ?

gregory-schiano: Also just saw that the SwiftService instance is create right after this

lib/charms/nginx_ingress_integrator/v0/ingress.py
arturo-seijas: This is the ingress library, directly pulled from charmhub, so we shouldn't be touching it here

lib/charms/nginx_ingress_integrator/v0/ingress.py
arturo-seijas: This is the ingress library, directly pulled from charmhub, so we shouldn't be touching it here

arturo-seijas (approved):
LGTM

tests/integration/test_core.py
weiiwang01: That's really convenient. Fixed, thanks!

lib/charms/nginx_ingress_integrator/v0/ingress.py
gregory-schiano: Erf, didn't realize it, then we should change this in the library itself

gregory-schiano (approved):
LGTM

PR #28, opened 2022-12-05
Add more integration tests

weiiwang01:
- add integration tests for multi-unit deployment and database from config
- reorganize test cases to speed up the integration difficulties
- fix a bug in the swift plugin reconciliation process in non-leader units


tests/integration/test_core.py
```diff
@@ -37,15 +38,7 @@ async def test_build_and_deploy(ops_test: pytest_operator.plugin.OpsTest, applic
     assert: building and deploying should success and status should be "blocked" since the
         database info hasn't been provided yet.
     """
-    my_charm = await ops_test.build_charm(".")
     assert ops_test.model
-    await ops_test.model.deploy(
-        my_charm,
-        resources={"wordpress-image": "localhost:32000/wordpress:test"},
```
jdkandersson, line 44: Maybe there is a wordpress charm? Or it can be attached as a resource?

.github/workflows/integration_test.yaml
```diff
@@ -46,3 +59,5 @@ jobs:
           --openid-password ${{ secrets.TEST_OPENID_PASSWORD }}
           --kube-config ${GITHUB_WORKSPACE}/kube-config
           --launchpad-team ${{ secrets.TEST_LAUNCHPAD_TEAM }}
+          -k ${{ matrix.module }}
+          ${{ matrix.args.args }}
```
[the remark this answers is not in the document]
weiiwang01: Fixed, thanks!

.github/workflows/integration_test.yaml
```diff
@@ -6,6 +6,19 @@ on:
 jobs:
   integration-tests:
     runs-on: ubuntu-latest
+    strategy:
```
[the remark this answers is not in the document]
weiiwang01: In addition to the parallelization on module, it need parallelization on test arguments too, which is not possible in `operator-workflow`. And adding that to `operator-workflow` might break backward compatibility.

tests/integration/test_error.py
```diff
@@ -0,0 +1,49 @@
...
+    act: provide incorrect database info via config.
+    assert: charm should be blocked by WordPress installation errors, instead of lacking
+        of database connection info.
+    """
+    # Database configuration can retry for up to 60 seconds before giving up and showing an error.
+    # Default wait_for_idle 15 seconds in ``app_config`` fixture is too short for incorrect
+    # db config.
+    assert ops_test.model
+    await ops_test.model.wait_for_idle(
+        idle_period=60, status=ops.model.BlockedStatus.name, apps=[application_name]  # type: ignore
```
amandahla, line 39: Does blocked status count as idle?

.github/workflows/integration_test.yaml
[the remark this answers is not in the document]
jdkandersson: Sounds like it would be worth trying, it would be quite handy to have that support on operator workflow

jdkandersson (dismissed):

tests/integration/conftest.py
```diff
@@ -330,8 +330,17 @@ def is_mysql_ready():
         build_and_deploy_wordpress(),
         wait_mysql_pod_ready(),
         ops_test.model.deploy("charmed-osm-mariadb-k8s", application_name="mariadb"),
-        ops_test.model.deploy(
-            "nginx-ingress-integrator", "ingress", trust=True, channel="edge", series="focal"
+        # temporary fix for the CharmHub problem
+        ops_test.juju(
+            "deploy",
```
jdkandersson, line 335: This and the `.deploy` call look identical?

tests/integration/conftest.py
```diff
@@ -261,3 +262,74 @@ def create_self_signed_tls_secret(host):
...
+            my_charm,
+            resources={"wordpress-image": "localhost:32000/wordpress:test"},
+            application_name=application_name,
+            series="jammy",
+            num_units=num_units,
+        )
+
+    kube_core_client.create_namespaced_pod(
+        namespace=ops_test.model_name,
+        body=kubernetes.client.V1Pod(
```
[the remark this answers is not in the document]
gregory-schiano: Could you add a comment to explain why we need a mysql deployed as a pod ?

tests/integration/test_core.py
```diff
@@ -60,53 +53,48 @@ async def test_build_and_deploy(ops_test: pytest_operator.plugin.OpsTest, applic
...
-        ), "unit status message should show detailed installation failure"
+    application = ops_test.model.applications[application_name]
+    await application.set_config(
+        {
+            "db_host": kube_core_client.read_namespaced_pod(
+                name="mysql", namespace=ops_test.model_name
+            ).status.pod_ip,
+            "db_name": "wordpress",
+            "db_user": "wordpress",
+            "db_password": "wordpress-password",
```
gregory-schiano, line 76: I'd use a constants for these parameters as they're also used when you create the mysql pod

tests/integration/conftest.py
jdkandersson: Ah, do you think that should be added to the comment as well? Just so we know what to look for in the future

jdkandersson (dismissed):

tests/integration/test_error.py
weiiwang01: Sorry, I put the replies in a comment so they were hold. I just released them.

tests/integration/conftest.py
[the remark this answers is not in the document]
weiiwang01: Added, thanks!

jdkandersson (dismissed):

jdkandersson (approved):

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM

PR #29, opened 2022-12-08
Add automated upgrade testing

weiiwang01:


src/charm.py
```diff
@@ -1142,26 +1159,62 @@ def _plugin_swift_reconciliation(self):
...
         ]
-        enable_swift = bool(swift_config_str.strip())
-        if not enable_swift:
+        swift_config = safe_load(swift_config_str)
+        if not swift_config:
+            return {}
+        if "url" in swift_config:
+            swift_url = swift_config["url"]
+            swift_url = re.sub("/wp-content/uploads/?$", "", swift_url)
+            swift_url = swift_url[: -(1 + len(swift_config.get("bucket", "")))]
```
gregory-schiano, line 1168: Probably a comment would be interesting here as this is not trivial logic

src/charm.py
```diff
@@ -1125,8 +1138,12 @@ def _apache_disable_config(self, conf_name):
         self._run_cli(["a2disconf", conf_name])
         self._start_server()
 
-    def _plugin_swift_reconciliation(self):
-        """Reconciliation process for swift object storage (openstack-objectstorage-k8s) plugin."""
+    def _swift_config(self):
+        """Load swift configuration from charm config.
```
gregory-schiano, line 1142: A bit of description concerning the legacy openstack configuration conversion would help the comprehension of this method

src/charm.py
```diff
@@ -1125,8 +1138,12 @@ def _apache_disable_config(self, conf_name):
...
+
+        Returns:
+            Swift configuration in dict.
+        """
         swift_config_str = self.model.config["wp_plugin_openstack-objectstorage_config"]
         swift_config_key = [
```
gregory-schiano, line 1148: If I understand well, there are the mandatory keys in swift_config ?
If so, I'd rename the variable to something more explicit like `mandatory_swift_config_key`

src/charm.py
```diff
@@ -1186,12 +1239,22 @@ def _plugin_reconciliation(self):
         and adjust plugin options for these three plugins according to charm config.
         """
         self._addon_reconciliation("plugin")
+        self._plugin_swift_reconciliation()
```
gregory-schiano, line 1242: Could you explain why you moved this call outside the `is_leader` ?

tests/integration/conftest.py
```diff
@@ -261,3 +290,166 @@ def create_self_signed_tls_secret(host):
 
     for secret in created_secrets:
         kube_core_client.delete_namespaced_secret(name=secret, namespace=namespace)
+
+
+@pytest.fixture(scope="module", name="deploy_and_wait_for_mysql_pod")
+def deploy_and_wait_for_mysql_pod_fixture(ops_test, kube_core_client):
+    """Return an async function that deploy and wait for a mysql pod ready in current namespace."""
```
gregory-schiano, line 297: I'd explain here why you need to deploy a "pod" manually, as while reading the code I was confused on why you'd need such fixture (until I saw the test)

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,257 @@
...
+    arrange: all charms and pods are deployed successfully.
+    act: enable some plugins and create posts and upload some images to the test WordPress blog.
+    assert: all operations finished without error.
+    """
+    namespace = ops_test.model_name
+
+    def get_wordpress_podspec_pod():
+        for pod in kube_core_client.list_namespaced_pod(namespace=namespace).items:
+            name = pod.metadata.name
+            if name.startswith("wordpress") and not name.endswith("-0"):
```
gregory-schiano, line 131: Can't you use labels to find the pod instead of its name ?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,257 @@
...
+            name=wordpress_pod,
+            namespace=namespace,
+            container="wordpress",
+            command=cmd,
+            stderr=True,
+            stdin=False,
+            stdout=True,
+            tty=False,
+        )
+        logger.info("Response: " + resp)
```
gregory-schiano: Wouldn't it be better to use pebble exec which is more the "Canonical" way ?

gregory-schiano: I'd even keep a bash script, copy it to the charm, and run it instead of doing multiple exec in a row

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+
+    charm = await ops_test.build_charm(".")
+    await ops_test.model.remove_application(application_name)
+
+    def wordpress_removed():
+        status = subprocess.check_output(
+            ["juju", "status", "-m", ops_test.model_name, "--format", "json"]
+        )
+        return application_name not in json.loads(status)["applications"]
```
gregory-schiano, line 204: ops_test doesn't provide other way to know when and application is effectively removed ?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,257 @@
...
+    for idx, unit_ip in enumerate(unit_ip_list):
+        await screenshot(f"http://{unit_ip}", screenshot_dir / f"wordpress-after-{idx}.png")
+        after_image = PIL.Image.open(screenshot_dir / f"wordpress-after-{idx}.png")
+        before_image = PIL.Image.open(screenshot_dir / f"wordpress-before-{idx}.png")
+        comparison_width = before_image.width + after_image.width
+        comparison_height = max(before_image.height, after_image.height)
+        comparison_image = PIL.Image.new("RGB", (comparison_width, comparison_height))
+        comparison_image.paste(before_image, (0, 0))
+        comparison_image.paste(after_image, (before_image.width, 0))
+        comparison_image.save(screenshot_dir / f"wordpress-comparison-{idx}.png")
```
gregory-schiano, line 249: If I understand well you creating an image that is an overlay of the 2 screenshots. Is it meant to be manually checked ? How do we ensure this is even checked ?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,257 @@
...
+    arrange: none.
+    act: deploy all required charms and kubernetes pods for tests.
+    assert: all charms and pods are deployed successfully.
+    """
+    assert ops_test.model
+
+    async def deploy_wordpress():
+        await ops_test.model.deploy(
+            "wordpress-k8s",
+            resources={"wordpress-image": "wordpresscharmers/wordpress:v5.9.4-20.04_edge"},
```
gregory-schiano, line 97: Why this version explicitly ?
This whole test seems to cover this very specific case, do we want to keep such specific test in the repo ?

src/charm.py
weiiwang01: The keys listed here is required for the WordPress OpenStack swift plugin, they all have to exist in the config for now, but I am planning to add some default values in the future.

src/charm.py
weiiwang01: Some steps of the `_plugin_swift_reconciliation` requires to run every unit, not just the leader unit, so move the `_plugin_swift_reconciliation` outside `is_leader`. And in `_plugin_swift_reconciliation`, update the code to run different processes in leader and non-leader units.

tests/integration/test_upgrade.py
weiiwang01: The current CharmHub version of the charm lacks some functionalities, after the CharmHub version of the charm deployed and configured, some manual operations is still needed to make the WordPress blog work. All the `exec` here is to run those manual operations automatically. This is just for testing and it will be removed once we published our version to CharmHub.

tests/integration/test_upgrade.py
weiiwang01: This is a side by side comparison, and for now it's manually checked. It's difficult for machine to check it since the charm upgrade may cause the theme upgrade, which will alter the appearance of the blog and that's completely normal. I have some plan in the future that post the comparison to the PR page as a comment, so the changes can be seen by code author and reviewers.

tests/integration/test_upgrade.py
weiiwang01: It's the latest version of the CharmHub version of the WordPress charm supported, and the new sidecar version is based on that. The entire `test_upgrade` is to mimic upgrading from one of our production deployment of the old WordPress charm. So some config is picked from the production config.

gregory-schiano: Ok let's keep it like this, but if in the future we want to test other version upgrade, we'll probably need to find a way to generalize these tests

tests/integration/test_upgrade.py
gregory-schiano: Ok I get, but wasn't pebble usable instead of kubernetes client for the exec method ?

src/charm.py
gregory-schiano: What about renaming the variable then ?

src/charm.py
weiiwang01: Add some comments, thanks!

src/charm.py
weiiwang01: Added, thanks!

src/charm.py
weiiwang01: Renamed to `required_swift_config_key`, thanks!

tests/integration/conftest.py
weiiwang01: Added, thanks!

tests/integration/test_upgrade.py
weiiwang01: The old version of the charm is downloaded and deployed from CharmHub. So, it's difficult to alter to code of the charm to run some pebble commands. Besides, all commands executed here can actually run everywhere with PHP installed, I am choosing to run inside the WordPress pod container because the PHP is already installed there, it's easier and time-saving to just run some commands there.

tests/integration/test_upgrade.py
weiiwang01: `ops_test` has `ops_test.model.get_status` but it's asynchronous, so it can't be used in `ops_test.model.block_until`.

tests/integration/test_upgrade.py
weiiwang01: Fixed, thanks!

gregory-schiano (commented):
The integration tests seems to fail due to the ingress unit

src/charm.py
```diff
@@ -1216,6 +1255,16 @@ def _plugin_reconciliation(self):
             self._plugin_openid_reconciliation()
 
     def _reconciliation(self, _event):
+        if not self.state.started:
```
arturo-seijas, line 1258: It is my understanding that `self._container().can_connect()` will always return false if `not self.state.started`. Why this additional check?

tests/integration/conftest.py
```diff
@@ -352,9 +373,32 @@ def is_mysql_ready():
...
+    deploy_and_wait_for_mysql_pod,
+):
+    """Deploy all required charms and kubernetes pods for tests."""
+    assert ops_test.model
+
+    async def build_and_deploy_wordpress():
+        my_charm = await ops_test.build_charm(".")
+        await ops_test.model.deploy(
+            my_charm,
+            resources={"wordpress-image": "localhost:32000/wordpress:test"},
```
arturo-seijas, line 393: It'd be good to parametrize this test so that the image can be passed as cmd arg. This would avoid coupling the test to the registry

src/charm.py
weiiwang01: Juju limits everything that happens before the `start` event to under 40 seconds (300 seconds for the newer version of juju). If exceeded, the container agent will be marked as unhealthy and restarted by Kubernetes. Since the reconciliation process may take some time, I am deferring everything until the `start` event happens.

tests/integration/conftest.py
```diff
@@ -282,63 +310,56 @@ def pod_db_password_fixture():
...
-    pod_db_user,
-    pod_db_password,
+@pytest.fixture(scope="module", name="deploy_and_wait_for_mysql_pod")
+def deploy_and_wait_for_mysql_pod_fixture(
+    ops_test, kube_core_client, pod_db_database, pod_db_user, pod_db_password
 ):
-    """Deploy all required charms and kubernetes pods for tests."""
-    assert ops_test.model
-    num_units = request.config.getoption("--num-units")
+    """Return an async function that deploy and wait for a mysql pod ready in current namespace.
```
arturo-seijas, line 317: Is it possible to drop support for this feature?

tests/integration/conftest.py
```diff
@@ -370,3 +414,73 @@ def is_mysql_ready():
         ),
     )
     await ops_test.model.wait_for_idle()
+
+
+@pytest.fixture(scope="module", name="test_image")
+def image_fixture():
```
arturo-seijas, line 420: I don't see where this fixture is used

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,259 @@
...
```
arturo-seijas, line 97: Same here, can we parametrize this?

tests/unit/test_charm.py
```diff
@@ -579,6 +600,54 @@ def test_team_map():
...
+    act: update legacy version of the wp_plugin_openstack-objectstorage_config configuration
+    assert: parsed swift configuration should update all legacy fields.
+    """
+    harness.set_can_connect(harness.model.unit.containers["wordpress"], True)
+    setup_replica_consensus()
+    harness.update_config(
+        {
+            "wp_plugin_openstack-objectstorage_config": json.dumps(
+                {
+                    "auth-url": "http://10.126.72.107/identity/v3",
```
arturo-seijas, line 618: Where does this IP come from?

tests/integration/conftest.py
weiiwang01: This is used in `tests.integration.test_upgrade.test_create_example_blog`.

tests/unit/test_charm.py
weiiwang01: It's just a random IP address from my local devstack server, the IP address here does not matter.

src/charm.py
arturo-seijas: Understood

tests/unit/test_charm.py
weiiwang01: I changed the IP address to `swift.test` to be self-explanatory.

.github/workflows/integration_test.yaml
```diff
@@ -19,7 +19,26 @@ jobs:
             value: "--num-units=3"
           - name: num_units=1, db_from_relation
             value: "--num-units=1"
+        exclude:
```
jdkandersson, line 22: Out of curiosity, what does this `exclude` do?

pyproject.toml
```diff
@@ -48,4 +48,4 @@ ignore_missing_imports = true
 allow_redefinition = true
 
 [tool.pylint.'MESSAGES CONTROL']
-disable = "too-few-public-methods,too-many-arguments,too-many-lines,line-too-long,fixme"
+disable = "too-few-public-methods,too-many-arguments,too-many-lines,line-too-long,too-many-locals,fixme,duplicate-code"
```
jdkandersson, line 51: What is the reason for adding these? Is it on test or source files?

src/charm.py
```diff
@@ -1149,7 +1165,34 @@ def _swift_config(self):
...
             return {}
-        for key in swift_config_key:
+        # legacy version of the WordPress charm accepts the ``url`` options
+        # here's an example of the ``url`` option:
+        # http://10.126.72.107:8080/v1/AUTH_fa8326b9fd4f405fb1c5eaafe988f5fd/WordPress/wp-content/uploads/
+        # which is a combination of swift url, container name, and object prefix
+        # the new WordPress charm will only take the swift url
+        # TODO: instead of user input, lookup swift url using swift client automatically
+        if "url" in swift_config:
+            swift_url = swift_config["url"]
```
jdkandersson, line 1175: Would it be worth putting the execution of the `if` statements into a separate method?

src/charm.py
```diff
@@ -1162,14 +1205,10 @@ def _config_swift_plugin(self, swift_config):
         Args:
             swift_config: swift configuration parsed from wp_plugin_openstack-objectstorage_config
                 config. Use :meth:`WordpressCharm._swift_config` to get the parsed swift config.
-
-        Raises:
```
jdkandersson, line 1166: Is it no longer possible for this exception ot occur? I couldn't spot it being removed in a change below

src/charm.pyi
```diff
@@ -63,17 +63,16 @@ class WordpressCharm(ops.charm.CharmBase):
 
     _WORDPRESS_DEFAULT_THEMES: List[str]
     _WORDPRESS_DEFAULT_PLUGINS: List[str]
-
+    def _on_start(self, _event: ops.charm.StartEvent): ...
```
jdkandersson, line 66: Should these define return types?

tests/conftest.py
```diff
@@ -35,3 +35,7 @@ def pytest_addoption(parser: pytest.Parser):
     parser.addoption("--test-db-from-config", action="store_true", default=False)
     # Number of WordPress units should the test deploy
     parser.addoption("--num-units", action="store", type=int, default=1)
+    # A directory to store screenshots generated by test_upgrade
+    parser.addoption("--screenshot-dir", action="store", default="")
+    # WordPress docker image built for the WordPress charm, default: localhost:32000/wordpress:test
```
jdkandersson, line 40: Is the default value needed in the comment? The code below seems to be quite readable in terms of the default value

tests/integration/conftest.py
```diff
@@ -107,13 +112,19 @@ def _get_plugin_list_from_ip(unit_ip):
...
     """Parse the openstack rc style configuration file from the --openstack-rc argument
 
-    Return a dictionary of environment variables and values.
+    Returns: a dictionary of environment variables and values, or None if --openstack-rc isn't
+        provided.
     """
     rc_file = request.config.getoption("--openstack-rc")
+    assert (
+        num_units == 1 or rc_file
+    ), "swift plugin is required for multi-unit deployment, please include an openstack configuration in the --openstack-rc parameter "
```
jdkandersson, line 125: This line looks like it is longer than 99 characters?

tests/integration/conftest.py
```diff
@@ -107,13 +112,19 @@ def _get_plugin_list_from_ip(unit_ip):
     return _get_plugin_list_from_ip
 
 
-@pytest.fixture
-def openstack_environment(request):
+@pytest.fixture(scope="module", name="openstack_environment")
```
jdkandersson, line 115: isn't the default `module`?

tests/integration/conftest.py
```diff
@@ -282,63 +316,56 @@ def pod_db_password_fixture():
...
-            resources={"wordpress-image": "localhost:32000/wordpress:test"},
-            application_name=application_name,
-            series="jammy",
-            num_units=num_units,
-        )
+    This is used for testing WordPress charm's capability of interacting with an external non-charm
+    MySQL database.
+    """
 
-    # create a pod to test the capability of the WordPress charm to interactive with an external
```
jdkandersson, line 309: Do you think this comment is still relevant?

tests/integration/conftest.py
```diff
@@ -370,3 +421,73 @@ def is_mysql_ready():
         ),
     )
     await ops_test.model.wait_for_idle()
+
+
+@pytest.fixture(scope="module", name="test_image")
+def image_fixture():
+    """A PNG image that can be used in tests."""
+    image_response = requests.get(
+        "https://s.w.org/style/images/about/WordPress-logotype-wmark.png", timeout=10
```
jdkandersson, line 430: Should we vendor this? As we move towards GitHub self-hosted runners HTTP access to the external world will be reduced and vending would also mean that if, for some reason, that image is removed, our tests don't break

tests/integration/conftest.py
```diff
@@ -370,3 +421,73 @@ def is_mysql_ready():
...
+    )
+    assert image_response.status_code == 200
+    return image_response.content
+
+
+@pytest.fixture(scope="module", name="swift_conn")
+def swift_conn_fixture(openstack_environment):
+    """Create a swift connection client."""
+    if openstack_environment is None:
+        return None
```
jdkandersson, line 440: Are we ok with returning `None`? If a test depends on this fixture but receives `None`, would you expect the test to still work?

tests/integration/conftest.py
```diff
@@ -370,3 +421,73 @@ def is_mysql_ready():
...
+            "project_name": openstack_environment["OS_PROJECT_NAME"],
+        },
+    )
+    return swift_conn
+
+
+@pytest.fixture(scope="module", name="swift_config")
+def swift_config_fixture(request: pytest.FixtureRequest, swift_conn, openstack_environment):
+    """Create a swift config dict that can be used for wp_plugin_openstack-objectstorage_config."""
+    if openstack_environment is None:
```
jdkandersson, line 458: Similar question as above

tests/integration/conftest.py
```diff
@@ -370,3 +421,73 @@ def is_mysql_ready():
...
+            "user_domain_name": openstack_environment["OS_USER_DOMAIN_ID"],
+            "project_domain_name": openstack_environment["OS_PROJECT_DOMAIN_ID"],
+            "project_name": openstack_environment["OS_PROJECT_NAME"],
+        },
+    )
+    return swift_conn
+
+
+@pytest.fixture(scope="module", name="swift_config")
+def swift_config_fixture(request: pytest.FixtureRequest, swift_conn, openstack_environment):
```
jdkandersson, line 456: Would type hints on the fixtures be valuable here?

tests/integration/conftest.py
```diff
@@ -370,3 +421,73 @@ def is_mysql_ready():
...
+            auth_version="3",
+            os_auth_url=openstack_environment["OS_AUTH_URL"],
+            os_username=openstack_environment["OS_USERNAME"],
+            os_password=openstack_environment["OS_PASSWORD"],
+            os_project_name=openstack_environment["OS_PROJECT_NAME"],
+            os_project_domain_name=openstack_environment["OS_PROJECT_DOMAIN_ID"],
+        )
+    )
+    container = f"wordpress_{request.module.__name__}"
+    print(f"Use container {container}")
```
jdkandersson, line 471: Is this intended to be there still?

tests/integration/test_core.py
```diff
@@ -126,61 +123,16 @@ async def test_openstack_object_storage_plugin(
         After openstack swift plugin activated, an image file uploaded to one unit through
         WordPress media uploader should be accessible from all units.
     """
+    if swift_config is None:
```
jdkandersson, line 126: Would it be possible to use this? https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-skipif

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,260 @@
...
+logger = logging.getLogger()
+
+POST_CONTENT = """\
+WordPress powers more than 39% of the web — a figure that rises every day.
+Everything from simple websites, to blogs, to complex portals and enterprise
+websites, and even applications, are built with WordPress. WordPress combines
+simplicity for users and publishers with under-the-hood complexity for
+developers. This makes it flexible while still being easy-to-use."""
+
+POST_CONTENT = POST_CONTENT.replace("\n", " ")
```
jdkandersson, line 34: What about moving this to `"...to-use."""".replace("\n", " ")`?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,260 @@
...
+
+POST_TITLE = "WordPress Post #1"
+
+POST_COMMENT = "I am a comment."
+
+
+async def screenshot(url, path):
+    """Create a screenshot of a website."""
+    async with async_playwright() as playwright:
+        browser = await playwright.chromium.launch()
```
jdkandersson, line 44: Would it be worth it to re-use a running chromium between tests? E.g., make this a fixture that cleans up chromium after the tests?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,260 @@
...
+        time.sleep(10)
+        await page.screenshot(full_page=True, path=path)
+        await browser.close()
+
+
+@pytest.fixture(scope="module", name="gen_upgrade_test_charm_config")
+def gen_upgrade_test_charm_config_fixture(ops_test, swift_config, kube_core_client):
+    """Create a function that generates charm config for upgrading tests."""
+    swift_url = swift_config["swift-url"]
+    swift_config = copy.copy(swift_config)
```
jdkandersson, line 56: is a shallow copy ok here? Otherwise there is also the `deepcopy` which copies not just the top level variables but does it recursively

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,260 @@
...
+            num_units=num_units,
+        )
+
+    await asyncio.gather(
+        deploy_wordpress(),
+        deploy_and_wait_for_mysql_pod(),
+        ops_test.run("playwright", "install", "chromium"),
+    )
+    await ops_test.model.applications[application_name].set_config(gen_upgrade_test_charm_config())
+    await ops_test.model.wait_for_idle(status=ops.model.ActiveStatus.name)  # type: ignore
```
jdkandersson, line 108: This looks like more of a fixture than a test?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,260 @@
...
+        host=unit_ip,
+        username="admin",
+        password=default_admin_password,
+        is_admin=True,
+    )
+    media_id = client.upload_media(filename="wordpress.png", content=test_image)["id"]
+    post = client.create_post(title=POST_TITLE, content=POST_CONTENT, featured_media=media_id)
+    client.create_comment(post_id=post["id"], post_link=post["link"], content=POST_COMMENT)
+    for idx, unit_ip in enumerate(unit_ip_list):
+        await screenshot(f"http://{unit_ip}", screenshot_dir / f"wordpress-before-{idx}.png")
```
jdkandersson, line 184: This also looks like a fixture?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,260 @@
...
+    await ops_test.model.block_until(wordpress_removed, wait_period=5, timeout=600)
+    await ops_test.model.deploy(
+        str(charm),
+        resources={"wordpress-image": "localhost:32000/wordpress:test"},
+        application_name=application_name,
+        series="jammy",
+        num_units=num_units,
+        config=gen_upgrade_test_charm_config(),
+    )
+    await ops_test.model.wait_for_idle(status=ops.model.ActiveStatus.name)  # type: ignore
```
jdkandersson, line 218: Also looks like a fixture?

tests/integration/test_upgrade.py
jdkandersson: Would it be worth it to include a comment explaining this in the code?

tests/unit/test_charm.py
```diff
@@ -579,6 +600,54 @@ def test_team_map():
...
+                    "remove-local-file": "0",
+                    "url": "http://swift.test:8080/v1/AUTH_fa8326b9fd4f405fb1c5eaafe988f5fd/wordpress_tests.integration.test_upgrade/wp-content/uploads/",
+                    "prefix": "wp-content/uploads/",
+                }
+            )
+        }
+    )
+    charm: WordpressCharm = typing.cast(WordpressCharm, harness.charm)
+    assert charm._swift_config() == {
+        "auth-url": "http://swift.test/identity/v3",
```
jdkandersson, line 636: Would it be worth it to define a variable to hold this configuration so that it doesn't have to be defined twice?

.github/workflows/integration_test.yaml
weiiwang01: Exclude some combinations created by matrix. Some of the test arguments are redundant for `test_error` and `test_upgrade`, exclude those cases to save some resources.

jdkandersson: Makes sense

src/charm.py
weiiwang01: I think this is removed when I merged the `sidecar` branch into this PR. I will add it back.

tests/integration/conftest.py
weiiwang01: Black does not break the string literal even if it's longer than 99 characters.

tests/integration/conftest.py
weiiwang01: The default scope is `function`.

tests/integration/conftest.py
weiiwang01: This is also caused by merging the `sidecar` branch into this PR. I will add this back too.

tests/integration/conftest.py
jdkandersson: ah yes, you have to do it manually in those cases. `pylint` should complain though, unless `line-too-long` is disabled 😉  which it is in this repo

tests/integration/conftest.py
jdkandersson: You are right

tests/integration/conftest.py
weiiwang01: I am a little bit concerned about the copyright and trademark issue if we include this image in our project.

tests/integration/conftest.py
weiiwang01: No, the test should fail as intended.

tests/integration/conftest.py
jdkandersson: Can we use another image or does it have to be this one? E.g., just pick something in the public domain?

tests/integration/conftest.py
weiiwang01: No, I will change this into `logger.info`, thanks!

tests/integration/test_core.py
weiiwang01: To skip the test or not depends on a fixture value. `pytest.mark.skipif` does not support that I think.

.github/workflows/integration_test.yaml
yanksyoon: Does the syntax for exclude need both name & value of the args to exclude to be repeated? Not familiar with exclude 😞

src/charm.pyi
weiiwang01: Added, thanks!

tests/conftest.py
weiiwang01: Removed, thanks!

tests/integration/conftest.py
weiiwang01: Yes, let's use a Canonical logo instead.

tests/integration/conftest.py
weiiwang01: Added, thanks!

tests/integration/test_upgrade.py
weiiwang01: `swift_config` is a `Dict[str, str]`, that's why shallow copy is okay here.

tests/integration/test_upgrade.py
weiiwang01: Here should have an assertion that three screenshots from different units should be identical, but it hasn't been implemented yet.

tests/integration/test_upgrade.py
weiiwang01: Here should have an assertion that three screenshots from different units should be identical, but it hasn't been implemented yet.

tests/integration/test_upgrade.py
weiiwang01: Added, thanks!

tests/unit/test_charm.py
weiiwang01: The swift configuration above is actually different from this one. Notice the `url` and `prefix`.

tests/integration/conftest.py
weiiwang01: Fixed, thanks!

tests/integration/test_upgrade.py
jdkandersson: Ok, I don't mind the way that looks but up to you

tests/integration/conftest.py
jdkandersson: That helps too, I was thinking about the arguments as well?

src/charm.py
weiiwang01: Added. thanks!

tests/integration/conftest.py
weiiwang01: Added. thanks!

tests/unit/test_charm.py
jdkandersson: ah, so maybe those things would just be updated before the comparison? That would make it clearer what the difference is as well

tests/integration/test_core.py
jdkandersson: Hmm, how often would this be skipped and what would be the reason for the skip? Is this if a parameter isn't supplied or something like that?

pyproject.toml
weiiwang01: `duplicate-code` is not needed anymore, I removed that. `too-many-locals` is in `tests/integration/test_upgrade.py`.

tests/integration/conftest.py
weiiwang01: Same as above.

tests/unit/test_charm.py
weiiwang01: Updated, thanks!

tests/integration/test_upgrade.py
weiiwang01: It will complicate the process, e.g. managing tabs opened by tests. It's not that slow to open a new browser.

tests/integration/conftest.py
weiiwang01: That has been added too.

tests/integration/test_core.py
weiiwang01: The WordPress charm support two mode: connecting to an external database using charm config and connecting to a charmed database using charm relation. The skip here is to enable integration tests to test both cases with different command line parameters.

.github/workflows/integration_test.yaml
weiiwang01: Since the `args` is an object (dict) in matrix, it has to be an object in `exclude` too

pyproject.toml
jdkandersson: Would it be better to specifically add it to the files where it needs to be disabled? E.g., just put `# pylint: disable=too-many-locals` at the top of the file and it will be disabled for the whole file. This way the check is disabled for all files

weiiwang01: Updated, thanks!

arturo-seijas (dismissed):
LGTM

arturo-seijas (dismissed):
LGTM

src/charm.py
```diff
@@ -812,14 +825,14 @@ def _addon_reconciliation(self, addon_type):
...
         uninstall_addons = current_installed_addons - desired_addons
         for addon in install_addons:
-            logger.debug("Install %s: %s", addon_type, repr(addon))
+            logger.info("Install %s: %s", addon_type, repr(addon))
             result = self._wp_addon_install(addon_type=addon_type, addon_name=addon)
             if not result.success:
                 raise exceptions.WordPressBlockedStatusException(
                     f"failed to install {addon_type} {repr(addon)}"
                 )
         for addon in uninstall_addons:
```
gtrkiller, line 834: Wouldn't it be convenient to execute the two for loops on a separate method?

src/charm.py
weiiwang01: This section will be removed in the future (see the TODO above), I think I will just leave it here for now.

tests/integration/test_upgrade.py
weiiwang01: Changed to a fixture, thanks!

tests/integration/test_upgrade.py
weiiwang01: Changed to a fixture, thanks!

tests/integration/test_upgrade.py
weiiwang01: Changed to a fixture, thanks!

arturo-seijas (dismissed):
LGTM

gregory-schiano (commented):
Some integration tests are still failing, I'll approve once green

src/charm.py
weiiwang01: It's just 13 lines of code for now, but I will definitely move that to a new method if there are more features added to these loops.

weiiwang01: Or do you mean that one method can handle both the installing and uninstalling logic to reduce duplicate code? I considered that and I actually did the same thing for `_addon_reconciliation` which combines the `_theme_reconciliation` and `_plugin_reconciliation`. But this two for-loops is not very long, so I didn't do that. As before, I will definitely that if more features are added to these loops.

gregory-schiano (approved):
LGTM

arturo-seijas (approved):
LGTM

PR #40, opened 2023-02-02
Add specific people to require review

jdkandersson:
We're going to experiment with requiring a review from specific people in addition to at least 2 overall reviews who have some expertise with a given repository. I'm not sure if this will work as expected since there are multiple `*` entries


merkata (dismissed):
Let's try it out. What is the end goal - having at least one approval from one of the specific engineers before merging or something similar?

arturo-seijas:
Are we sure we want this? This will impact delivery. Moreover, the three needed reviewers are in the same time zone

gregory-schiano:
> Let's try it out. What is the end goal - having at least one approval from one of the specific engineers before merging or something similar?

I agree, what do we want to achieve / avoid here ?
I'm not sure it'll work anyway as it requires 2 approval from people in the codeowners, unless you also changed the approval rules ?

jdkandersson:
> > Let's try it out. What is the end goal - having at least one approval from one of the specific engineers before merging or something similar?
> 
> I agree, what do we want to achieve / avoid here ? I'm not sure it'll work anyway as it requires 2 approval from people in the codeowners, unless you also changed the approval rules ?

The reason we have this is to require approvals from a SME on the charm. So who is on the secondary list of the codeowners would depend charm by charm, in the case of Wordpress it happens to be most of the APAC team. And yes, it does slow down delivery. This PR basically states that we want the SME of the charm to review changes.

I'm not yet convinced that this will technically work. The branch protection already requires a review from the codeowners, I just don't know how specifically GitHub works because the two lines of the file have the same glob. I think the way it works is that a review from each line that matches is required.

nrobinaubertin:
> I think the way it works is that a review from each line that matches is required.

No, "the last matching pattern takes the most precedence", cf: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

jdkandersson:
> > I think the way it works is that a review from each line that matches is required.
> 
> No, "the last matching pattern takes the most precedence", cf: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

Thanks for sharing that, looks like the proposed approach would result in all the PRs just going to the specific people, probably not what we want. An alternative could be to make the codeowners the SMEs and add a GitHub action on PR creation that adds the team as reviewers as well

jdkandersson:
I also tried to request the reviewers via a workflow, the problem is that you can't request a team because the GITHUB_TOKEN doesn't have read permissions to read the teams

PR #42, opened 2023-02-08
fix: Open ID auth

yanksyoon:
This PR is to fix the failing test for OpenID plugin integration tests. It is simplified from #41 to unblock the non-passing CI first before committing to the bigger tests refactor.

The problems are in:
1. PHP version which is incompatible with wordpress-launchpad-integration, hence the Dockerfile's base image downgrade(to support php ver 7)
2. Updated Ubuntu One website, hence the regex fix

### DIFF
--- a/wordpress.Dockerfile
+++ b/wordpress.Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:20.04
 
 ARG VERSION=5.9.3
 ENV APACHE_CONFDIR=/etc/apache2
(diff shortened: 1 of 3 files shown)


wordpress.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:20.04
```
jdkandersson, line 1: What is the reason for the downgrade please?

jdkandersson (approved):

yhaliaw (approved):

PR #43, opened 2023-02-13
feat: COS integration

yanksyoon:
This PR adds a new COS integration to wordpress-k8s-operator charm.

The elements of COS added are prometheus, loki and grafana.

Prometheus monitors apache used to serve wordpress.
Loki scrapes logs from apache server, including access log and error log. Container logs are also preserved in addition to the appending to log files for loki.
Grafana contains a pre-made dashboard template to monitor the apache server.

Loki screenshot
image

Grafana screenshot (updated 2022/02/13)
image

Note for reviewers:
Configuration files are mostly default values. Added lines of code from the configuration values have been marked with a comment on the PR.


files/000-default.conf
```diff
@@ -0,0 +1,31 @@
...
+	DocumentRoot /var/www/html
+
+	# Available loglevels: trace8, ..., trace1, debug, info, notice, warn,
+	# error, crit, alert, emerg.
+	# It is also possible to configure the loglevel for particular
+	# modules, e.g.
+	#LogLevel info ssl:warn
+
+	ErrorLog "|$ tee {APACHE_LOG_DIR}/error.log"
+    CustomLog "|$ tee ${APACHE_LOG_DIR}/access.log" combined
```
yanksyoon, line 21: Apache logs are piped through tee and log files to enable both container log output and file append for loki.

files/apache2.conf
```diff
@@ -0,0 +1,230 @@
...
+<Directory /var/www/>
+        Options Indexes FollowSymLinks
+        AllowOverride None
+        Require all granted
+</Directory>
+
+<Location "/server-status">
+    SetHandler server-status
+    Allow from localhost
+</Location>
```
yanksyoon, line 179: /server-status route allow configuration is added for apache-prometheus-exporter to be able to expose metrics for prometheus scrape job.

files/docker-php.conf
```diff
@@ -17,5 +17,6 @@ DirectoryIndex index.php index.html
 	RewriteRule ^index\.php$ - [L]
 	RewriteCond %{REQUEST_FILENAME} !-f
 	RewriteCond %{REQUEST_FILENAME} !-d
+	RewriteCond %{REQUEST_URI} !server-status
```
yanksyoon, line 20: To allow apache to handle server-status route rather than forwarding to wordpress server routes.

yanksyoon (commented):
Leaving comments for configuration files that are mostly untouched.

src/grafana_dashboards/wordpress.json
```diff
@@ -0,0 +1,425 @@
...
+        "y": 1
+      },
+      "id": 9,
+      "options": {
+        "code": {
+          "language": "plaintext",
+          "showLineNumbers": false,
+          "showMiniMap": false
+        },
+        "content": "# Wordpress Operator for Juju\n\n- Github [URL](https://github.com/canonical/wordpress-k8s-operator)\n\n- For more information, visit wordpress [charmnub](https://charmhub.io/wordpress-k8s)\n\n",
```
mthaddon, line 105: s/charmnub/charmhub/

yanksyoon: Great catch thank you!

mthaddon:
I'm not sure that "uptime", "apache version" and "apache workers" are such important bits of information that we'd want to surface them in this way on the dashboard. Is there a list somewhere of the metrics we have available so we can decide what should be shown where?

src/grafana_dashboards/wordpress.json
```diff
@@ -0,0 +1,425 @@
...
+    "list": []
+  },
+  "time": {
+    "from": "now-6h",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "Wordpress Operator Overview",
+  "uid": "yQ2wKmT4z",
```
arturo-seijas, line 422: Iirc the uid is generated when the dashboard is injected. There's no need to specify it

yanksyoon:
@mthaddon Sure, below are the lists of metrics exposed from apache-prometheus-exporter.

Current improvements(upcoming push) include:
- Graph: `promhttp_metric_handler_requests_in_flight`
- Gauge: `promhttp_metric_handler_requests_total`
- Graph: `apache_access_total`
- Gauge: `apache_load`
- Gauge: `apache_scoreboard`

```
# HELP apache_accesses_total Current total apache accesses (*)
# TYPE apache_accesses_total counter
apache_accesses_total 1060
# HELP apache_cpu_time_ms_total Apache CPU time
# TYPE apache_cpu_time_ms_total counter
apache_cpu_time_ms_total{type="system"} 4960
apache_cpu_time_ms_total{type="user"} 23820
# HELP apache_cpuload The current percentage CPU used by each worker and in total by all workers combined (*)
# TYPE apache_cpuload gauge
apache_cpuload 0.814144
# HELP apache_duration_ms_total Total duration of all registered requests in ms
# TYPE apache_duration_ms_total counter
apache_duration_ms_total 73319
# HELP apache_exporter_build_info A metric with a constant '1' value labeled by version, revision, branch, and goversion from which apache_exporter was built.
# TYPE apache_exporter_build_info gauge
apache_exporter_build_info{branch="HEAD",goversion="go1.16.10",revision="f4fd9dd7e9672fda120a3085f224431550baf2a7",version="0.11.0"} 1
# HELP apache_generation Apache restart generation
# TYPE apache_generation gauge
apache_generation{type="config"} 1
apache_generation{type="mpm"} 0
# HELP apache_info Apache version information
# TYPE apache_info gauge
apache_info{mpm="prefork",version="Apache/2.4.41 (Ubuntu) OpenSSL/1.1.1f"} 1
# HELP apache_load Apache server load
# TYPE apache_load gauge
apache_load{interval="15min"} 3.3
apache_load{interval="1min"} 2.07
apache_load{interval="5min"} 2.43
# HELP apache_scoreboard Apache scoreboard statuses
# TYPE apache_scoreboard gauge
apache_scoreboard{state="closing"} 0
apache_scoreboard{state="dns"} 0
apache_scoreboard{state="graceful_stop"} 0
apache_scoreboard{state="idle"} 6
apache_scoreboard{state="idle_cleanup"} 0
apache_scoreboard{state="keepalive"} 0
apache_scoreboard{state="logging"} 0
apache_scoreboard{state="open_slot"} 143
apache_scoreboard{state="read"} 0
apache_scoreboard{state="reply"} 1
apache_scoreboard{state="startup"} 0
# HELP apache_sent_kilobytes_total Current total kbytes sent (*)
# TYPE apache_sent_kilobytes_total counter
apache_sent_kilobytes_total 3780
# HELP apache_up Could the apache server be reached
# TYPE apache_up gauge
apache_up 1
# HELP apache_uptime_seconds_total Current uptime in seconds (*)
# TYPE apache_uptime_seconds_total counter
apache_uptime_seconds_total 3535
# HELP apache_version Apache server version
# TYPE apache_version gauge
apache_version 2.04041
# HELP apache_workers Apache worker statuses
# TYPE apache_workers gauge
apache_workers{state="busy"} 1
apache_workers{state="idle"} 6
# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 1.68
# HELP process_max_fds Maximum number of open file descriptors.
# TYPE process_max_fds gauge
process_max_fds 65536
# HELP process_open_fds Number of open file descriptors.
# TYPE process_open_fds gauge
process_open_fds 10
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 1.6052224e+07
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1.6762743514e+09
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 7.33540352e+08
# HELP process_virtual_memory_max_bytes Maximum amount of virtual memory available in bytes.
# TYPE process_virtual_memory_max_bytes gauge
process_virtual_memory_max_bytes 1.8446744073709552e+19
# HELP promhttp_metric_handler_requests_in_flight Current number of scrapes being served.
# TYPE promhttp_metric_handler_requests_in_flight gauge
promhttp_metric_handler_requests_in_flight 1
# HELP promhttp_metric_handler_requests_total Total number of scrapes by HTTP status code.
# TYPE promhttp_metric_handler_requests_total counter
promhttp_metric_handler_requests_total{code="200"} 353
promhttp_metric_handler_requests_total{code="500"} 0
promhttp_metric_handler_requests_total{code="503"} 0
```

Note that go related (go_gc_*) are from prometheus go client's default profiling which is not related to WordPress apache server(below).
```
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 2.7648e-05
go_gc_duration_seconds{quantile="0.25"} 3.9714e-05
go_gc_duration_seconds{quantile="0.5"} 5.8567e-05
go_gc_duration_seconds{quantile="0.75"} 0.00011677
go_gc_duration_seconds{quantile="1"} 0.012443423
go_gc_duration_seconds_sum 0.03104215
go_gc_duration_seconds_count 63
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 10
# HELP go_info Information about the Go environment.
# TYPE go_info gauge
go_info{version="go1.16.10"} 1
# HELP go_memstats_alloc_bytes Number of bytes allocated and still in use.
# TYPE go_memstats_alloc_bytes gauge
go_memstats_alloc_bytes 5.275424e+06
# HELP go_memstats_alloc_bytes_total Total number of bytes allocated, even if freed.
# TYPE go_memstats_alloc_bytes_total counter
go_memstats_alloc_bytes_total 2.1709904e+08
# HELP go_memstats_buck_hash_sys_bytes Number of bytes used by the profiling bucket hash table.
# TYPE go_memstats_buck_hash_sys_bytes gauge
go_memstats_buck_hash_sys_bytes 1.462217e+06
# HELP go_memstats_frees_total Total number of frees.
# TYPE go_memstats_frees_total counter
go_memstats_frees_total 595998
# HELP go_memstats_gc_cpu_fraction The fraction of this program's available CPU time used by the GC since the program started.
# TYPE go_memstats_gc_cpu_fraction gauge
go_memstats_gc_cpu_fraction 1.511275573844487e-05
# HELP go_memstats_gc_sys_bytes Number of bytes used for garbage collection system metadata.
# TYPE go_memstats_gc_sys_bytes gauge
go_memstats_gc_sys_bytes 5.075784e+06
# HELP go_memstats_heap_alloc_bytes Number of heap bytes allocated and still in use.
# TYPE go_memstats_heap_alloc_bytes gauge
go_memstats_heap_alloc_bytes 5.275424e+06
# HELP go_memstats_heap_idle_bytes Number of heap bytes waiting to be used.
# TYPE go_memstats_heap_idle_bytes gauge
go_memstats_heap_idle_bytes 5.8925056e+07
# HELP go_memstats_heap_inuse_bytes Number of heap bytes that are in use.
# TYPE go_memstats_heap_inuse_bytes gauge
go_memstats_heap_inuse_bytes 7.00416e+06
# HELP go_memstats_heap_objects Number of allocated objects.
# TYPE go_memstats_heap_objects gauge
go_memstats_heap_objects 4228
# HELP go_memstats_heap_released_bytes Number of heap bytes released to OS.
# TYPE go_memstats_heap_released_bytes gauge
go_memstats_heap_released_bytes 5.5230464e+07
# HELP go_memstats_heap_sys_bytes Number of heap bytes obtained from system.
# TYPE go_memstats_heap_sys_bytes gauge
go_memstats_heap_sys_bytes 6.5929216e+07
# HELP go_memstats_last_gc_time_seconds Number of seconds since 1970 of last garbage collection.
# TYPE go_memstats_last_gc_time_seconds gauge
go_memstats_last_gc_time_seconds 1.6762778918424115e+09
# HELP go_memstats_lookups_total Total number of pointer lookups.
# TYPE go_memstats_lookups_total counter
go_memstats_lookups_total 0
# HELP go_memstats_mallocs_total Total number of mallocs.
# TYPE go_memstats_mallocs_total counter
go_memstats_mallocs_total 600226
# HELP go_memstats_mcache_inuse_bytes Number of bytes in use by mcache structures.
# TYPE go_memstats_mcache_inuse_bytes gauge
go_memstats_mcache_inuse_bytes 14400
# HELP go_memstats_mcache_sys_bytes Number of bytes used for mcache structures obtained from system.
# TYPE go_memstats_mcache_sys_bytes gauge
go_memstats_mcache_sys_bytes 16384
# HELP go_memstats_mspan_inuse_bytes Number of bytes in use by mspan structures.
# TYPE go_memstats_mspan_inuse_bytes gauge
go_memstats_mspan_inuse_bytes 194480
# HELP go_memstats_mspan_sys_bytes Number of bytes used for mspan structures obtained from system.
# TYPE go_memstats_mspan_sys_bytes gauge
go_memstats_mspan_sys_bytes 212992
# HELP go_memstats_next_gc_bytes Number of heap bytes when next garbage collection will take place.
# TYPE go_memstats_next_gc_bytes gauge
go_memstats_next_gc_bytes 9.977584e+06
# HELP go_memstats_other_sys_bytes Number of bytes used for other system allocations.
# TYPE go_memstats_other_sys_bytes gauge
go_memstats_other_sys_bytes 1.966327e+06
# HELP go_memstats_stack_inuse_bytes Number of bytes in use by the stack allocator.
# TYPE go_memstats_stack_inuse_bytes gauge
go_memstats_stack_inuse_bytes 1.179648e+06
# HELP go_memstats_stack_sys_bytes Number of bytes obtained from system for stack allocator.
# TYPE go_memstats_stack_sys_bytes gauge
go_memstats_stack_sys_bytes 1.179648e+06
# HELP go_memstats_sys_bytes Number of bytes obtained from system.
# TYPE go_memstats_sys_bytes gauge
go_memstats_sys_bytes 7.5842568e+07
# HELP go_threads Number of OS threads created.
# TYPE go_threads gauge
go_threads 18
```

src/grafana_dashboards/wordpress.json
yanksyoon: I see, thank you for that! Updating the tests w/ removed uid :)

mthaddon:
In terms of what metrics to display and how on the dashboard it may be worth breaking that out to a separate discussion/spec document. If that's going to slow down the landing of this PR I think it's fine to go ahead as is and come back later with dashboard updates as long as that's relatively easy to do, if you prefer.

My initial reaction would be to ideally have something like "Requests/second (to Wordpress itself)", "CPU load" and "Resident Memory usage" as the main things we're calling out in the dashboard (in place of "uptime", "apache version" and "apache workers"), but it would be good to get feedback from others on this as well. Being able to see some kind of metric that shows apache error codes over the last 5 minutes (or some similar interval) to see how many 5XXs there are (for instance) in comparison to "normal" would be good as well.

arturo-seijas (commented):
Can you add the units of each metrics in the dashboard as part of the tittle (seconds, requests,etc)?

arturo-seijas (dismissed):
LGTM

files/000-default.conf
weiiwang01: Isn't repeating the `CustomLog` creating multiple log stream? To quote the [apache document](https://httpd.apache.org/docs/2.4/mod/mod_log_config.html):

> Three directives are provided by this module: [TransferLog](https://httpd.apache.org/docs/2.4/mod/mod_log_config.html#transferlog) to create a log file, [LogFormat](https://httpd.apache.org/docs/2.4/mod/mod_log_config.html#logformat) to set a custom format, and [CustomLog](https://httpd.apache.org/docs/2.4/mod/mod_log_config.html#customlog) to define a log file and format in one step. The TransferLog and CustomLog directives can be used multiple times in each server to cause each request to be logged to multiple files.

So this could be update to something like this, maybe?

```
CustomLog "${APACHE_LOG_DIR}/access.log" combined
CustomLog /dev/stdout combined
```

tests/integration/test_core.py
```diff
@@ -540,3 +545,107 @@ async def test_openid_plugin(
...
+                "user": "admin",
+                "password": password,
+            },
+        ).raise_for_status()
+        datasources = sess.get(f"http://{unit.address}:3000/api/datasources", timeout=10).json()
+        datasource_types = tuple(datasource["type"] for datasource in datasources)
+        assert "loki" in datasource_types
+        assert "prometheus" in datasource_types
+        dashboards = sess.get(
+            f"http://{unit.address}:3000/api/search?query=Wordpress Operator Overview", timeout=10
```
weiiwang01, line 649: `requests` has a `params` argument for HTTP query string

```python
sess.get(
    f"http://{unit.address}:3000/api/search",
    params={"query": "Wordpress Operator Overview"}, 
    timeout=10
)
```

yanksyoon: Thanks for catching that!

files/apache2.conf
jdkandersson: Do you think it would be worth making this a comment in the file? The PR comment will be difficult to track down once the PR is merged

jdkandersson: Similar comment for other comments 😉

files/000-default.conf
yanksyoon: I thought pipe operation would be consistent with ErrorLog but it does seem better to use native log stream directives. Updated, thank you for the comment!

files/apache2.conf
yanksyoon: It does seem very necessary :)
Added comments of the format
# Date of change
# Reason for change
for all the configuration files, thank you for the feedback!

src/charm.py
```diff
@@ -1375,10 +1409,30 @@ def _reconciliation(self, _event: EventBase) -> None:
...
+        else:
+            self.unit.status = WaitingStatus("Waiting for pebble")
+
+    def _on_apache_prometheus_exporter_pebble_ready(self, event: PebbleReadyEvent):
+        """Configure and start apache prometheus exporter.
+
+        Args:
+            event: Event triggering the handler.
+        """
+        if not event.workload:
```
jdkandersson, line 1427: Out of curiosity, do you know under which scenarios this would occur? I'm guessing the `workload` is the prometheus exporter in this case?

tests/integration/test_core.py
```diff
@@ -540,3 +545,111 @@ async def test_openid_plugin(
...
+    assert: prometheus metrics endpoint for prometheus is active and prometheus has active scrape
+        targets.
+    """
+    assert ops_test.model
+    prometheus: Application = await ops_test.model.deploy(
+        "prometheus-k8s", channel="stable", trust=True
+    )
+
+    await ops_test.model.relate(application_name, prometheus.name)
+    await ops_test.model.wait_for_idle(apps=[application_name, prometheus.name], status="active")
```
jdkandersson, line 567: I'm not sure, does `apps` have to be a list? Or could it also be a tuple given it is a constant?

tests/integration/test_core.py
```diff
@@ -540,3 +545,111 @@ async def test_openid_plugin(
...
+
+    for unit_ip in unit_ip_list:
+        requests.get(
+            f"http://{unit_ip}:{APACHE_PROMETHEUS_SCRAPE_PORT}", timeout=10
+        ).raise_for_status()
```
jdkandersson, line 572: Would it be better to assert on the response code? I'm guessing the intent here is that the test should fail if something other than a `2XX` response is returned?

tests/integration/test_core.py
jdkandersson: Similar comment for any other constant collections

tests/integration/test_core.py
```diff
@@ -540,3 +545,111 @@ async def test_openid_plugin(
...
+    status: FullStatus = await ops_test.model.get_status(filters=[prometheus.name])
+    for unit in status.applications[prometheus.name].units.values():
+        query_targets = requests.get(
+            f"http://{unit.address}:9090/api/v1/targets", timeout=10
+        ).json()
+        assert len(query_targets["data"]["activeTargets"])
+
+
+async def test_loki_integration(
```
jdkandersson, line 581: I'm not as familiar with COS, does this test assume that the `test_prometheus_integration` was run before this one?

tests/integration/test_core.py
```diff
@@ -540,3 +545,111 @@ async def test_openid_plugin(
...
+    act: prometheus charm joins relation
+    assert: prometheus metrics endpoint for prometheus is active and prometheus has active scrape
+        targets.
+    """
+    assert ops_test.model
+    prometheus: Application = await ops_test.model.deploy(
+        "prometheus-k8s", channel="stable", trust=True
+    )
+
+    await ops_test.model.relate(application_name, prometheus.name)
```
jdkandersson, line 566: Should this relation be removed after the end of the test?

jdkandersson: Similar comment for any other tests adding integrations

tests/integration/test_core.py
```diff
@@ -540,3 +545,111 @@ async def test_openid_plugin(
...
+    application_name: str,
+    kube_core_client: kubernetes.client.CoreV1Api,
+):
+    """
+    arrange: after WordPress charm has been deployed and relations established.
+    act: loki charm joins relation
+    assert: loki joins relation successfully, logs are being output to container and to files for
+        loki to scrape.
+    """
+    assert ops_test.model
```
jdkandersson, line 592: An alternative to having this on every test is to create a fixture that exposes the `ops_test.model` so that the assert only needs to be done once

tests/integration/test_core.py
```diff
@@ -540,3 +545,111 @@ async def test_openid_plugin(
...
+    loki: Application = await ops_test.model.deploy("loki-k8s", channel="stable", trust=True)
+
+    await ops_test.model.relate(application_name, loki.name)
+    await ops_test.model.wait_for_idle(apps=[application_name, loki.name], status="active")
+
+    status: FullStatus = await ops_test.model.get_status(filters=[loki.name])
+    for unit in status.applications[loki.name].units.values():
+        series = requests.get(f"http://{unit.address}:3100/loki/api/v1/series", timeout=10).json()
+        log_files = tuple(series_data["filename"] for series_data in series["data"])
```
jdkandersson, line 601: Should this be a set given the following lines check for whether certain elements are in the collection?

tests/integration/test_core.py
```diff
@@ -540,3 +545,111 @@ async def test_openid_plugin(
...
+    """
+    arrange: after WordPress charm has been deployed and relations established among cos.
+    act: grafana charm joins relation
+    assert: grafana wordpress dashboard can be found
+    """
+    assert ops_test.model
+    grafana: Application = await ops_test.model.deploy("grafana-k8s", trust=True)
+
+    await ops_test.model.relate(application_name, grafana.name)
+    await ops_test.model.relate("prometheus-k8s:grafana-source", f"{grafana.name}:grafana-source")
```
jdkandersson, line 629: This does make it look like the other integrations are required, should they be fixtures instead so that these tests can be run in isolation rather than requiring them to be run in a certain order?

tests/integration/test_core.py
```diff
@@ -540,3 +545,111 @@ async def test_openid_plugin(
...
+        sess = requests.session()
+        sess.post(
+            f"http://{unit.address}:3000/login",
+            json={
+                "user": "admin",
+                "password": password,
+            },
+        ).raise_for_status()
+        datasources = sess.get(f"http://{unit.address}:3000/api/datasources", timeout=10).json()
+        datasource_types = tuple(datasource["type"] for datasource in datasources)
```
jdkandersson, line 647: Similar to above, should this be a set given the next lines check for elements within the collection?

jdkandersson (commented):
Thanks for putting this together, looks like the tests are thorough. There seem to be some comments on the Dockerfile, should we add an issue to fix those items?

tests/integration/test_core.py
yanksyoon: It does require the previous tests such as database relation & prometheus relation tests to run before.
This was deliberately done to match the current repository's testing style, but in hindsight, it does seem that the new tests can be better written to run independently and leave the previous tests to be refactored with another issue.
Will push the patches soon. Thank you for the comment :)

yanksyoon (commented):
Applying the fixes soon, thank you for the comment!

tests/integration/test_core.py
yanksyoon: It does seem like the `juju.Model` strictly requires it to be a list[str]
```
# from juju/model.py:L2626
if apps is not None and (not isinstance(apps, list) or
                                 any(not isinstance(o, str)
                                     for o in apps)):
            raise JujuError(f'Expected a List[str] for apps, given {apps}')
```

arturo-seijas (dismissed):
LGTM

yanksyoon:
Currently the tests are unstable, not sure if Github Actions being unstable is further worsening the situation. Committing some safeguard measures(wait for idle) to hopefully make the integration tests pass.

tests/integration/conftest.py
```diff
@@ -10,36 +10,51 @@
...
 
 
+@pytest.fixture(scope="module", name="model")
+def model_module_scope_fixture(ops_test: OpsTest) -> Model:
+    """Get current valid model created for integraion testing with module scope."""
+    assert ops_test.model
+    return ops_test.model
+
+
+@pytest.fixture(scope="function", name="func_model")
```
jdkandersson, line 37: Are both of these needed? Shouldn't one of these be enough? I'm guessing the module scope fixture can also be used where a function scope fixture would otherwise be used?

tests/integration/conftest.py
```diff
@@ -49,21 +64,17 @@ def fixture_application_name():
 
 
 @pytest_asyncio.fixture(scope="module", name="get_default_admin_password")
-async def fixture_get_default_admin_password(
-    ops_test: pytest_operator.plugin.OpsTest, application_name
-):
+async def fixture_get_default_admin_password(model: Model, application_name: str):
```
jdkandersson, line 67: much nicer!

tests/integration/test_core.py
```diff
@@ -558,28 +561,32 @@ async def test_prometheus_integration(
     assert: prometheus metrics endpoint for prometheus is active and prometheus has active scrape
         targets.
     """
-    assert ops_test.model
-    prometheus: Application = await ops_test.model.deploy(
-        "prometheus-k8s", channel="stable", trust=True
+    await prometheus.relate(
```
jdkandersson, line 564: Should this also be in the/ another fixture? Similar for the other relations, especially since it looks like the relations are being cleaned up

tests/integration/conftest.py
yanksyoon: Pytest complains about the scope being incompatible when using module scoped fixture in function scoped test/fixture, would there be another way around it?

tests/integration/test_core.py
yanksyoon: Changing to fixture with `yield`s to allow cleanups in fixture and not impact the test. Thank you for the comment, TIL!

src/charm.py
jdkandersson: An alternative would be to go into blocked state or whatever is appropriate for this event failing

tests/integration/test_core.py
yanksyoon: Refactored all fixtures to include application relation. Thank you for the feedback :)

tests/integration/test_core.py
yanksyoon: Relation removal after a test included in the fixtures now. Thanks for noticing that!

tests/integration/test_core.py
yanksyoon: Refactored to assert for status rather than raising. Much clearer what this code is trying to test now. Thank you for the feedback!

src/charm.py
yanksyoon: Updated using BlockedStatus rather than custom status with indicator saying `Internal Error, pebble container not found`. This case shouldn't really be happening since this event is derived from `WorkloadEvent` which should have container. 
```
class WorkloadEvent(HookEvent):
def __init__(self, handle: 'Handle', workload: 'Container'):
        super().__init__(handle)

        self.workload = workload
```
Thank you for the comment!

jdkandersson: Perhaps we should raise a bug for the upstream library to fix the type hint if it can't actually be None?

jdkandersson (dismissed):

src/charm.py
yanksyoon: Will do :)

mthaddon:
Thx for the dashboard updates. I think the "Apache workers" values are for "busy" (1) and "idle" (6) workers? Is there some way of making that clearer in the dashboard? Currently I think most people would see this as 16 workers and not be aware of what status they're in.

yanksyoon:
@mthaddon Sounds great, applied the same changes w/ regex renaming.
Updated dashboard screenshot:
image

jdkandersson (approved):

arturo-seijas (approved):
LGTM

PR #50, opened 2023-02-27
chore: ci run minial integration test

yanksyoon:
This PR removes the matrix run of test-and-publish-charm action which runs on multiple combinations and thus also publishes multiple instances of wordpress-k8s charm.

Since the local integration_test workflow runs the full suite of integration tests(test with secrets, combinations of tests using matrices), running the full suite of tests in test-and-publish-charm action is not needed.
The tests swapped in place for the matrix test in test-and-publish-charm is the minimal core integration test suite for the sake of running the test and building docker images.


amandahla (approved):
LGTM

jdkandersson (approved):

jdkandersson:
That is the reason why we had some many versions published

PR #52, opened 2023-02-28
docs: reference documentation

yanksyoon:
This PR implements reference documentation for wordpress-k8s charm using the [Diataxis framework](https://diataxis.fr/reference/).

Out of scope of this PR: other parts of documentation(tutorial, how-to guide, explanation). Later PRs will address the remaining documentations.


docs/reference/configurations.md
```diff
@@ -0,0 +1,81 @@
...
+  the WordPress role) - commas separate multiple pairs.
+
+  Example format:
+
+  "site-sysadmins=administrator,site-editors=editor,site-executives=editor"
+
+- _wp_plugin_openstack_objectstorage_config_: YAML dictionary with keys named after WordPress
+  settings and the desired values. Please note that the settings will be reset to values provided
+  every time hooks run. Below are available values of the configuration yaml.
+
```
weiiwang01, line 55: Maybe we should mention that the openstack object storage plugin is a must for multi-unit deployment?

docs/reference/plugins.md
```diff
@@ -0,0 +1,81 @@
+# Plugins
+
+By default, the following WordPress plugins are installed with the latest version during the OCI
+image build time. Hence, if WordPress is deployed at a later stage, there may be version
+differences between the plugins across the WordPress units.
```
weiiwang01, line 5: The discrepancy of plugins between units will only happen if user install additional plugins through the `plugin` configuration. The default plugins came with the docker image will always be the same across all units. But plugins defined by the `plugins` configuration will be installed when ever the plugin is not persisted in the pod. So the plugin defined by the `plugins` configuration can have a different installation time.

docs/reference/actions.md
```diff
@@ -0,0 +1,10 @@
+# Actions
+
+- _get-initial-password_: Retrieve auto-generated initial password for accessing WordPress admin
+  account. A new random password of length 32 characters is generated each time a new leader is
```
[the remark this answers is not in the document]
yanksyoon: Right, I think i should also mention that if the password is set on the `inital_settings` configuration, the get-initial-password action doesn't return the right password. Thank you for fact checking!

docs/reference/configurations.md
yanksyoon: Sounds good, added!

docs/reference/configurations.md
```diff
@@ -0,0 +1,81 @@
...
+  Optionally you can also provide
+
+  ```yaml
+  admin_password: <secret> # auto generated if not set
+  blog_public: False # by default blogs are public
+  ```
+
+  If _admin_password_ is not provided it will be automatically generated and stored on the operator
+  pod in the /root directory.
```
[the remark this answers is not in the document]
yanksyoon: Updated! Thank you for noticing!

docs/reference/actions.md
```diff
@@ -0,0 +1,10 @@
...
+  elected, when the charm is upgraded or when rotate-wordpress-secrets action is run.
+
+- _rotate-wordpress-secrets_: Rotate all WordPress secrets, used for WordPress to encrypt
```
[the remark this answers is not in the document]
yanksyoon: Added further description of invalidating sessions/users :) thank you!

docs/reference/configurations.md
```diff
@@ -0,0 +1,81 @@
...
+  ```yaml
+  user_name: admin_username
+  admin_email: name@example.com
+  ```
+
+  Optionally you can also provide
+
+  ```yaml
+  admin_password: <secret> # auto generated if not set
+  blog_public: False # by default blogs are public
```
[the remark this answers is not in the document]
yanksyoon: Updated with changes in `config.yaml` as well. Thank you for your comment :)

docs/reference/plugins.md
yanksyoon: Updated the documentation to note the information on runtime installed plugins. Thank you for that!

docs/reference/actions.md
```diff
@@ -0,0 +1,12 @@
+# Actions
+
+- _get-initial-password_: Retrieve auto-generated initial password for accessing WordPress admin
```
arturo-seijas, line 3: Action name in the actions.yaml is `get-initial-password`

docs/reference/configurations.md
```diff
@@ -0,0 +1,81 @@
...
+    to enable object versioning in Swift bucket. Read more about object versioning from the source
+    documentation.
+  password: (string) OpenStack password. Example: openstack_secret_password
+  permissions: (string)
+  region: (string) OpenStack region
+  remove-local-file: (string) string value ‘1’ or ‘0’ denoting true, false respectively on whether
+    to remove local file. Example: 0
+  serve-from-swift: (string) string value ‘1’ or ‘0’ denoting true, false respectively on whether
+    to serve the contents file directly from swift. Example: 1
+  swift-url: (string) OpenStack Swift URL. example: http://10.100.115.2:8080/v1
```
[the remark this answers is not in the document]
yanksyoon: I'm currently checking the configurations as i'm writing the how-to guide, but this was obtained from the source code `https://git.launchpad.net/~canonical-sysadmins/wordpress/+git/openstack-objectstorage-k8s/tree/classes/swift.php#n495`. Let me prioritise confirming openstack related parameters first and will post an update. :)

docs/reference/actions.md
```diff
@@ -0,0 +1,12 @@
...
+  account. The password is set once during deployment. If the wordpress-k8s charm is configured
+  with `initial_settings` parameters containing `admin_password`, this action has no effect.
+
+- _rotate-wordpress-secrets_: Rotate all WordPress secrets, used for WordPress to encrypt
```
arturo-seijas, line 7: Action name in actions.yaml is `rotate-wordpress-secrets`

docs/reference/actions.md
```diff
@@ -0,0 +1,12 @@
...
+  information. The secret fields are used to enhance security by encrypting WordPress data. The
+  following secrets are rotated: auth_key, secure_auth_key, logged_in_key, nonce_key, auth_salt,
```
arturo-seijas, line 9: Can you order the secrets alphabetically?

docs/reference/actions.md
yanksyoon: May I know if you mean the markdown syntax for italics _action-name_ is confusing to read? I could update that with *action-name* or get rid of the itatlics formatting altogether if it reads better :)
Thank you for your comment!

docs/reference/actions.md
```diff
@@ -0,0 +1,12 @@
...
+  secure_auth_salt, logged_in_salt, nonce_salt
+  The purpose of this action is to invalidate user sessions under security breach circumstances
```
arturo-seijas, line 11: Is the purpose to rotate the secrets or to invalidate? Perhpas this paragraph would be better phrased with something like
```
this action WILL invalidate user sessions forcing the users to log in again. This might be useful under security breach circumstances
```
Also, keep in mind that when documenting, the simpler the grammar, the better

docs/reference/configurations.md
```diff
@@ -0,0 +1,82 @@
+# Configurations
+
+- _blog_hostname_: Hostname for accessing WordPress, if ingress relation is active. Defaults to the
```
arturo-seijas, line 3: Same here, put the config names as they will be used in cmd line, that is, as they appear in config.yaml

docs/reference/actions.md
arturo-seijas: Nvm me, I just noticed those are italics

docs/reference/actions.md
yanksyoon: Resolving :) (same reason as above)

docs/reference/actions.md
yanksyoon: Reordered alphabetically :) Thank you for your comment.

docs/reference/actions.md
yanksyoon: Rephrased to simpler phrases:
```
 Invalidate user sessions by rotating the following secrets:
  auth_key, auth_salt, logged_in_key, logged_in_salt, nonce_key, nonce_salt, secure_auth_key,
  secure_auth_salt.
  Users will be forced to log in again. This might be useful under security breach circumstances.
```
Thank you for your suggestion :)

docs/reference/plugins.md
```diff
@@ -0,0 +1,82 @@
+# Plugins
+
+By default, the following WordPress plugins are installed with the latest version during the OCI
+image build time. If the plugins are installed during runtime with
+`juju config wordpress-k8s plugins=...`, the plugin will also be installed to it's latest version
+by default and may cause version differences between pods.
+Due to the clustered nature of the WordPress charm, installing plugins through UI has been disabled
+and can only be installed through the plugins configuration. Please see Configurations section for
```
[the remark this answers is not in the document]
yanksyoon: Linked to the configuration file! Thank you for the suggestion :)

arturo-seijas (dismissed):
LGTM

docs/reference/configurations.md
```diff
@@ -0,0 +1,87 @@
...
```
jdkandersson, line 3: What is the reason for not putting this on the configuration options? On charmhub, there is a default page for the configurations: https://charmhub.io/wordpress-k8s/configure?channel=edge. The same comment applies for actions

docs/reference/integrations.md
```diff
@@ -0,0 +1,66 @@
...
+### db
+
+_Interface_: mysql  
+_Supported charms_: [charmed-osm-mariadb-k8s](https://charmhub.io/mysql),
+[mysql-k8s](https://charmhub.io/mysql-k8s)
+
+Database integration is a required relation for the wordpress-k8s charm to supply structured data
+storage for WordPress. It is recommended to use a juju native integration that provides a mysql
+interface by providing necessary configuration details to wordpress-k8s charm. See
+[configuration](configuration.md) for more specific parameters.
```
jdkandersson, line 12: Will this link [configuration](configuration.md) work on charmhub?

docs/reference/integrations.md
```diff
@@ -0,0 +1,66 @@
...
+
+### db
+
+_Interface_: mysql  
+_Supported charms_: [charmed-osm-mariadb-k8s](https://charmhub.io/mysql),
+[mysql-k8s](https://charmhub.io/mysql-k8s)
+
+Database integration is a required relation for the wordpress-k8s charm to supply structured data
+storage for WordPress. It is recommended to use a juju native integration that provides a mysql
+interface by providing necessary configuration details to wordpress-k8s charm. See
```
jdkandersson, line 11: Would it be worth including more information about which are the necessary configuration options in the case of using an integration?

jdkandersson: What about an example for how to configure the integration?

docs/reference/plugins.md
```diff
@@ -0,0 +1,82 @@
+# Plugins
+
+By default, the following WordPress plugins are installed with the latest version during the OCI
+image build time. If the plugins are installed during runtime with
+`juju config wordpress-k8s plugins=...`, the plugin will also be installed to it's latest version
+by default and may cause version differences between pods.
+Due to the clustered nature of the WordPress charm, installing plugins through UI has been disabled
+and can only be installed through the plugins configuration. Please see
+[Configurations](configurations.md) section for more information.
```
jdkandersson, line 9: Similar to earlier note, I'm guessing this reference won't work on charmhub

docs/reference/configurations.md
yanksyoon: What do you think about removing the `configurations.md` and `actions.md` files altogether and just updating the `actions.yaml` & `config.yaml`? The way users will interact with the documentation is through the sidebar and if the desired action is to guide them towards correct tabs, I think it may be better to not have these two sections at all.
With reference to Indico charm, when the users to land on [reference/actions](https://charmhub.io/indico/docs/reference-actions), the only content is See Actions with a navlink to the actions tab which I think is not the best UX.

docs/reference/configurations.md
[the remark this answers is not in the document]
yanksyoon: Tested all the parameters, `hidpi-images` configuration works with WP Retina 2x plugin which isn't installed by default. Hence, I've decided to remove the configuration option.
Other parameters (permissions) have also been tested and been removed since they don't work.
Thank you for the comment, resolving :)

docs/reference/integrations.md
yanksyoon: Added `mysql-interface-user` and `mysql-interface-database` parameter configuration for mysql interface.

On the second note about example on how to configure the integration, I think examples are better suited in how-to guide or in the tutorial, which should be ready for review soon :)

docs/reference/plugins.md
yanksyoon: Updated to the configure tab link: https://charmhub.io/wordpress-k8s/configure?channel=edge
Thank you for the comment!

docs/reference/integrations.md
yanksyoon: Updated to the configure tab link: https://charmhub.io/wordpress-k8s/configure?channel=edge
Thank you for the comment!

docs/reference/configurations.md
yanksyoon: Resolved to adding a reference link to the according tabs as per discussion today. Thanks for the comment!

jdkandersson: As discussed, we will use the Indico method of having a document with just a link so that there is a navigation table entry until `upload-charm-docs` supports external links preperly

docs/reference/integrations.md
jdkandersson: Once it is on stable it might be confusing to get a link to edge

yanksyoon: Would it be better if the last documentation PR includes a cleanup of all the link pointing to edge so that the edge documentation can be easily confirmed with navigation before publishing to stable?
Otherwise, I'm happy to remove the edge channel link :)

jdkandersson: Sounds like a reasonable approach, although it would be good to get to stable sooner rather than later

weiiwang01 (dismissed):

jdkandersson (approved):

weiiwang01 (approved):

PR #60, opened 2023-03-13
Check /proc/mounts for storage attachment

weiiwang01:
Use /proc/mounts to check if a storage volume is attached to the container. 

The Juju `*-storage-attached` event is only fired once and is not re-emitted during a charm upgrade, which can cause the charm to get stuck waiting for storage.


src/charm.py
```diff
@@ -183,7 +175,7 @@ def __init__(self, *args, **kwargs):
 
         self.framework.observe(self.on.leader_elected, self._setup_replica_data)
         self.framework.observe(self.on.start, self._on_start)
-        self.framework.observe(self.on.uploads_storage_attached, self._on_storage_attached)
+        self.framework.observe(self.on.uploads_storage_attached, self._reconciliation)
```
arturo-seijas, line 178: Is this event triggered before the config change? If so, we can probably remove this listener given that the callback is the same
`self.framework.observe(self.on.config_changed, self._reconciliation)`

weiiwang01: I am not sure if the firing order of `*-storage-attached` event is guaranteed, it might fired after the `config-changed`. I just add a callback here to make sure the charm doesn't stuck waiting for storage.

arturo-seijas (dismissed):
LGTM

yanksyoon (commented):
I'm wondering if we could benefit from writing a test describing the case when this actually happens in the update_charm test. Looks good otherwise!

src/charm.py
```diff
@@ -1389,27 +1377,39 @@ def _plugin_reconciliation(self) -> None:
...
+        """Check if the upload storage mounted in the wordpress container.
+
+        Returns:
+            True if the storage "upload" is attached to the container.
+        """
+        container = self._container()
+        if not container.can_connect():
+            return False
+        mount_info: str = container.pull("/proc/mounts").read()
+        return "/var/www/html/wp-content/uploads" in mount_info
```
yhaliaw, line 1390: Would it be better to parse the content of `/proc/mounts` so we can check for a exact match?
Since `/proc/mounts` should have a well defined format, and this way of checking can have false positive in theory.

jdkandersson (commented):
I don't see any unit/ integration tests that catch this bug? It would be good to have that so that we can prevent any regressions

src/charm.py
weiiwang01: The problem of parsing the `/proc/mounts` is that it can be quite complex, as file names in Linux can contain whitespaces and newlines.

jdkandersson:
> > I don't see any unit/ integration tests that catch this bug? It would be good to have that so that we can prevent any regressions
> 
> I should be checked by the upgrade test, but the upgrade test is occupied by the upgrade from podspec version to sidecar version. The upgrade from sidecar version to sidecar version is not tested.

Should we add another upgrade test to check for the upgrade from sidecar to sidecar?

src/charm.py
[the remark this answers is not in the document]
amandahla: Sorry, I wasn't clear. If the directory "/var/www/html/wp-content/uploads/abc" is mounted but "/var/www/html/wp-content/uploads/" is empty, would that be ok?

weiiwang01:
> > > I don't see any unit/ integration tests that catch this bug? It would be good to have that so that we can prevent any regressions
> > 
> > 
> > I should be checked by the upgrade test, but the upgrade test is occupied by the upgrade from podspec version to sidecar version. The upgrade from sidecar version to sidecar version is not tested.
> 
> Should we add another upgrade test to check for the upgrade from sidecar to sidecar?

As, we will soon deprecate the podspec version of the charm, we can convert the current upgrade test to test the upgrade from sidecar to sidecar. If we add the test now, it requires us to create a test instead of convert the existing one.

jdkandersson:
> As, we will soon deprecate the podspec version of the charm, we can convert the current upgrade test to test the upgrade from sidecar to sidecar. If we add the test now, it requires us to create a test instead of convert the existing one.

Let's discuss in the next stand up

src/charm.py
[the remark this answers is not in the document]
weiiwang01: Yes, it is possible to achieve that. However, to ensure 100% accuracy, parsing the /proc/mounts file is necessary, which can be challenging due to various edge cases such as filenames containing white spaces and new line characters.

The mount points within the container should be managed by the metadata.yaml file. Therefore, I do not believe there is a high chance of false positives occurring in this case.

.github/workflows/integration_test.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
         --kube-config ${GITHUB_WORKSPACE}/kube-config \
         --screenshot-dir /tmp \
         ${{ matrix.args.value }}
-      modules: '["test_core", "test_error", "test_upgrade"]'
+      modules: '["test_core", "test_error", "test_upgrade", "test_new_upgrade"]'
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

weiiwang01:
> > Use /proc/mounts to check if a storage volume is attached to the container.
> > The Juju `*-storage-attached` event is only fired once and is not re-emitted during a charm upgrade, which can cause the charm to get stuck waiting for storage.
> 
> Isn't this a bug ? According to the event documentation: https://juju.is/docs/sdk/storage-name-storage-attached-event `The event is emitted after a storage volume has been attached to the charm’s host machine or container.`
> 
> When you do an upgrade, a new container is provisioned and therefore storage is attached to it, so I'd expect the event to be triggered.
> 
> This PR could be good as a workaround, but this definitely needs to be discussed with the Ops / Juju team so see what is the expected behavior

It really depends, juju attach a storage by modifying the stateful set. When the container is recreated in the event of upgrading, juju does not modifying the stateful set. It's understandable to not emit the `*-storage-attached` event in that case. Also, for machine charm, the storage attachment is a one time thing also. The storage is not detached and re-attached during the upgrade for machine charms. That's what I understand.

gregory-schiano:
> > > Use /proc/mounts to check if a storage volume is attached to the container.
> > > The Juju `*-storage-attached` event is only fired once and is not re-emitted during a charm upgrade, which can cause the charm to get stuck waiting for storage.
> > 
> > 
> > Isn't this a bug ? According to the event documentation: https://juju.is/docs/sdk/storage-name-storage-attached-event `The event is emitted after a storage volume has been attached to the charm’s host machine or container.`
> > When you do an upgrade, a new container is provisioned and therefore storage is attached to it, so I'd expect the event to be triggered.
> > This PR could be good as a workaround, but this definitely needs to be discussed with the Ops / Juju team so see what is the expected behavior
> 
> It really depends, juju attach a storage by modifying the stateful set. When the container is recreated in the event of upgrading, juju does not modifying the stateful set. It's understandable to not emit the `*-storage-attached` event in that case. Also, for machine charm, the storage attachment is a one time thing also. The storage is not detached and re-attached during the upgrade for machine charms. That's what I understand.

Ok I get it, so the issue is more on us using the storedState to store a flag.
Then why don't we use the recommended ways to get rid of the storedState:
- Writing a file in the said storage, and checking if that file exists
- Using the peer relation databag

This would avoid the trick of parsing the mounts

weiiwang01:
> > > > Use /proc/mounts to check if a storage volume is attached to the container.
> > > > The Juju `*-storage-attached` event is only fired once and is not re-emitted during a charm upgrade, which can cause the charm to get stuck waiting for storage.
> > > 
> > > 
> > > Isn't this a bug ? According to the event documentation: https://juju.is/docs/sdk/storage-name-storage-attached-event `The event is emitted after a storage volume has been attached to the charm’s host machine or container.`
> > > When you do an upgrade, a new container is provisioned and therefore storage is attached to it, so I'd expect the event to be triggered.
> > > This PR could be good as a workaround, but this definitely needs to be discussed with the Ops / Juju team so see what is the expected behavior
> > 
> > 
> > It really depends, juju attach a storage by modifying the stateful set. When the container is recreated in the event of upgrading, juju does not modifying the stateful set. It's understandable to not emit the `*-storage-attached` event in that case. Also, for machine charm, the storage attachment is a one time thing also. The storage is not detached and re-attached during the upgrade for machine charms. That's what I understand.
> 
> Ok I get it, so the issue is more on us using the storedState to store a flag. Then why don't we use the recommended ways to get rid of the storedState:
> 
> * Writing a file in the said storage, and checking if that file exists
> * Using the peer relation databag
> 
> This would avoid the trick of parsing the mounts

Yes, I think using the peer relation is plausible. But I think the approach proposed in this pull request is actually stateless thus much simpler than the peer relation in term of implementation.

jdkandersson (approved):

arturo-seijas (approved):
LGTM

gregory-schiano:
> > > > > Use /proc/mounts to check if a storage volume is attached to the container.
> > > > > The Juju `*-storage-attached` event is only fired once and is not re-emitted during a charm upgrade, which can cause the charm to get stuck waiting for storage.
> > > > 
> > > > 
> > > > Isn't this a bug ? According to the event documentation: https://juju.is/docs/sdk/storage-name-storage-attached-event `The event is emitted after a storage volume has been attached to the charm’s host machine or container.`
> > > > When you do an upgrade, a new container is provisioned and therefore storage is attached to it, so I'd expect the event to be triggered.
> > > > This PR could be good as a workaround, but this definitely needs to be discussed with the Ops / Juju team so see what is the expected behavior
> > > 
> > > 
> > > It really depends, juju attach a storage by modifying the stateful set. When the container is recreated in the event of upgrading, juju does not modifying the stateful set. It's understandable to not emit the `*-storage-attached` event in that case. Also, for machine charm, the storage attachment is a one time thing also. The storage is not detached and re-attached during the upgrade for machine charms. That's what I understand.
> > 
> > 
> > Ok I get it, so the issue is more on us using the storedState to store a flag. Then why don't we use the recommended ways to get rid of the storedState:
> > 
> > * Writing a file in the said storage, and checking if that file exists
> > * Using the peer relation databag
> > 
> > This would avoid the trick of parsing the mounts
> 
> Yes, I think using the peer relation is plausible. But I think the approach proposed in this pull request is actually stateless thus much simpler than the peer relation in term of implementation.

Well it's more following the convention that were suggested in the "Charm engineering" meeting when they mentioned that StoredState is getting deprecated and should be removed from charms.

I agree that your approach is totally stateless and probably better, but I don't agree this is simpler.
On top of that, I need to check but there could be cases where /proc or /proc/mounts is not available in the container due to securityContext or capabilities.
Last but not least, we're going towards unprivileged containers (which is a best practice in terms of security) for workload and juju agent, and I'm also not sure the mounts information will be available in such cases

weiiwang01:
> > > > > > Use /proc/mounts to check if a storage volume is attached to the container.
> > > > > > The Juju `*-storage-attached` event is only fired once and is not re-emitted during a charm upgrade, which can cause the charm to get stuck waiting for storage.
> > > > > 
> > > > > 
> > > > > Isn't this a bug ? According to the event documentation: https://juju.is/docs/sdk/storage-name-storage-attached-event `The event is emitted after a storage volume has been attached to the charm’s host machine or container.`
> > > > > When you do an upgrade, a new container is provisioned and therefore storage is attached to it, so I'd expect the event to be triggered.
> > > > > This PR could be good as a workaround, but this definitely needs to be discussed with the Ops / Juju team so see what is the expected behavior
> > > > 
> > > > 
> > > > It really depends, juju attach a storage by modifying the stateful set. When the container is recreated in the event of upgrading, juju does not modifying the stateful set. It's understandable to not emit the `*-storage-attached` event in that case. Also, for machine charm, the storage attachment is a one time thing also. The storage is not detached and re-attached during the upgrade for machine charms. That's what I understand.
> > > 
> > > 
> > > Ok I get it, so the issue is more on us using the storedState to store a flag. Then why don't we use the recommended ways to get rid of the storedState:
> > > 
> > > * Writing a file in the said storage, and checking if that file exists
> > > * Using the peer relation databag
> > > 
> > > This would avoid the trick of parsing the mounts
> > 
> > 
> > Yes, I think using the peer relation is plausible. But I think the approach proposed in this pull request is actually stateless thus much simpler than the peer relation in term of implementation.
> 
> Well it's more following the convention that were suggested in the "Charm engineering" meeting when they mentioned that StoredState is getting deprecated and should be removed from charms.
> 
> I agree that your approach is totally stateless and probably better, but I don't agree this is simpler. On top of that, I need to check but there could be cases where /proc or /proc/mounts is not available in the container due to securityContext or capabilities. Last but not least, we're going towards unprivileged containers (which is a best practice in terms of security) for workload and juju agent, and I'm also not sure the mounts information will be available in such cases

Yes, the WordPress charm is in the process of removing the `StoreState`, I think once we use the new mysql charm library from the data platform, the `StoreState` can be removed.

For now, I think we can just use the `/proc/mounts` to check the storage attachment. Even if the security context is added to juju, I think the charm owner will have the capability to customize the security context. In terms of unprivileged container, `/prox/mounts` does not require any elevated privilege to read. Also, is pebble gonna work in unprivileged mode?

gregory-schiano:
> > > > > > > Use /proc/mounts to check if a storage volume is attached to the container.
> > > > > > > The Juju `*-storage-attached` event is only fired once and is not re-emitted during a charm upgrade, which can cause the charm to get stuck waiting for storage.
> > > > > > 
> > > > > > 
> > > > > > Isn't this a bug ? According to the event documentation: https://juju.is/docs/sdk/storage-name-storage-attached-event `The event is emitted after a storage volume has been attached to the charm’s host machine or container.`
> > > > > > When you do an upgrade, a new container is provisioned and therefore storage is attached to it, so I'd expect the event to be triggered.
> > > > > > This PR could be good as a workaround, but this definitely needs to be discussed with the Ops / Juju team so see what is the expected behavior
> > > > > 
> > > > > 
> > > > > It really depends, juju attach a storage by modifying the stateful set. When the container is recreated in the event of upgrading, juju does not modifying the stateful set. It's understandable to not emit the `*-storage-attached` event in that case. Also, for machine charm, the storage attachment is a one time thing also. The storage is not detached and re-attached during the upgrade for machine charms. That's what I understand.
> > > > 
> > > > 
> > > > Ok I get it, so the issue is more on us using the storedState to store a flag. Then why don't we use the recommended ways to get rid of the storedState:
> > > > 
> > > > * Writing a file in the said storage, and checking if that file exists
> > > > * Using the peer relation databag
> > > > 
> > > > This would avoid the trick of parsing the mounts
> > > 
> > > 
> > > Yes, I think using the peer relation is plausible. But I think the approach proposed in this pull request is actually stateless thus much simpler than the peer relation in term of implementation.
> > 
> > 
> > Well it's more following the convention that were suggested in the "Charm engineering" meeting when they mentioned that StoredState is getting deprecated and should be removed from charms.
> > I agree that your approach is totally stateless and probably better, but I don't agree this is simpler. On top of that, I need to check but there could be cases where /proc or /proc/mounts is not available in the container due to securityContext or capabilities. Last but not least, we're going towards unprivileged containers (which is a best practice in terms of security) for workload and juju agent, and I'm also not sure the mounts information will be available in such cases
> 
> Yes, the WordPress charm is in the process of removing the `StoreState`, I think once we use the new mysql charm library from the data platform, the `StoreState` can be removed.
> 
> For now, I think we can just use the `/proc/mounts` to check the storage attachment. Even if the security context is added to juju, I think the charm owner will have the capability to customize the security context. In terms of unprivileged container, `/prox/mounts` does not require any elevated privilege to read. Also, is pebble gonna work in unprivileged mode?

Yes there are 2 ongoing discussions:
- Unprivileged workload (pebble and the underlying service)
- Unprivileged juju agent

Best practices in terms of containers is to avoid AT ALL any process running as root in a container.
Concerning the mounts, not all mounts are available to users, so depending how the storage is mounted, is could be invisible to a non-root user.

Fine to go on with this, and we'll see on a longer term when the "unprivileged discussions" will be concluded and implemented.

PR #64, opened 2023-03-28
feat: DatabaseRequires library integration

yanksyoon:
This PR marks old mysql relation for [deprecation](https://github.com/canonical/mysql-k8s-operator/blob/538635ab0ab023bccc06d6c5822397dd79d4db32/src/relations/mysql.py#L164) and integrates with the new `data_interfaces` [library](https://github.com/canonical/data-platform-libs).

The next sets of improvements that can come from this PR include TLS connection to the database from the WordPress application. Currently, the TLS support in mysql-k8s charm is [under progress](https://github.com/canonical/mysql-k8s-operator/blob/538635ab0ab023bccc06d6c5822397dd79d4db32/src/relations/mysql_provider.py#L221).

Minor additions: library upgrade


src/charm.py
```diff
@@ -370,11 +383,50 @@ def _on_relation_database_changed(self, event: MySQLDatabaseChangedEvent) -> Non
...
+
+    def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
+        """Store database connection configuration details to charm state.
+
+        Args:
+            event: Event emitted when a new database is created for use on this relation,
+            propagated from relation_changed event under the hood.
+        """
+        self.state.relation_db_host = self._parse_database_endpoints(endpoints=event.endpoints)
+        self.state.relation_db_name = event.database
```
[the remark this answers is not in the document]
yanksyoon: I wanted to keep the existing method of reconciliation as is for now and have a major refactor on the next ticket. But it seems that it's necessary to refactor this piece of code now since using the same state will cause bugs when both db relations are related and one departs.
Thank you for the comment :)

src/constants.py
```diff
@@ -0,0 +1,7 @@
+# Copyright 2023 Canonical Ltd.
```
arturo-seijas, line 1: IMHO the approach of defining constants in a dedicated file breaks encapsulation. Is there any reason why these are not defined in `charm.py`?

yanksyoon: I took the idea from [mysql-k8s-operator repo](https://github.com/canonical/mysql-k8s-operator/blob/d83c4f3c6a05d0ec688412f2e497eb9c94a5b603/src/constants.py), it seemed better to separate the common constants out into a separate file so that later when other relations get put into a separate file there will be no cyclic dependency problem.
My intention for the refactor was to allow grounding for each relation to be separated it it's own file to better separate out code for each of the relation in it's own respective file.
Since this refactor should probably go through a spec, I'm fine with reverting this particular change and moving the constants in `charm.py`. What do you think?

arturo-seijas: I'd go with reverting since there's currently no motivation for keeping these constants here. Perhaps in the future it will be a good solution.

yanksyoon: Done :)

arturo-seijas (dismissed):
LGTM

metadata.yaml
```diff
@@ -52,6 +52,8 @@ requires:
   db:
     interface: mysql
     limit: 1
```
weiiwang01, line 54: Should we consider drop the support for the legacy mysql relation?

src/charm.py
```diff
@@ -582,26 +597,98 @@ def _wp_is_installed(self):
...
+
+        Returns:
+            Hostname of database running on port 3306. None if no endpoints are provided.
+            Note that WordPress will throw MySQL Error when supplying a port with the hostname
+            (i.e. host:port).
+        """
+        if not endpoints:
+            return None
+        urls = endpoints.split(",")
+        return urls[0].split(":")[0]
```
weiiwang01, line 654: The default mysql port for WordPress is 3306, should we throw an error here if the port provided by mysql charm is not listen on 3306?

metadata.yaml
yanksyoon: I was thinking about deprecating it gradually like the nginx-ingress-integrator, perhaps setting a timeline/notice for deprecation?

src/charm.py
```diff
@@ -582,26 +597,98 @@ def _wp_is_installed(self):
...
-    def _current_effective_db_info(self):
+    @property
+    def _config_database_config(self) -> types_.DatabaseConfig | None:
+        """Database configuration details from charm config.
+
+        Returns:
+            Database configuration required to establish database connection.
+            None if not exists.
+        """
+        if any(
```
jdkandersson, line 608: This would be `True` if only one of the items in the list is actually present, is that as intended?

src/charm.py
```diff
@@ -582,26 +597,98 @@ def _wp_is_installed(self):
...
+            self.model.config.get(key) for key in ["db_host", "db_name", "db_user", "db_password"]
```
jdkandersson, line 609: Since this collection is constant, should it use a tuple?

src/charm.py
```diff
@@ -582,26 +597,98 @@ def _wp_is_installed(self):
...
+                username=self.model.config.get("db_user"),
+                password=self.model.config.get("db_password"),
+            )
+        return None
+
+    @property
+    def _db_relation_database_config(self) -> types_.DatabaseConfig | None:
+        """Database configuration details from stored state.
+
+        Since the legacy db relation is per-unit basis, the relation data must be stored in charm
```
jdkandersson, line 623: To help me understand, where in this function is the interaction with `StoredState`?

src/charm.py
```diff
@@ -582,26 +597,98 @@ def _wp_is_installed(self):
...
+            endpoints: Comma separated endpoints of format host:port
+
+        Returns:
+            Hostname of database running on port 3306. None if no endpoints are provided.
+            Note that WordPress will throw MySQL Error when supplying a port with the hostname
+            (i.e. host:port).
+        """
+        if not endpoints:
+            return None
+        urls = endpoints.split(",")
```
jdkandersson, line 653: It looks like we only use one of the endpoints, what could be the reason if there are multiple endpoints? Perhaps this is intended for load balancing purposes which is being undone by always picking the first?

src/charm.py
```diff
@@ -610,14 +697,13 @@ def _test_database_connectivity(self):
...
         """
-        db_info = self._current_effective_db_info()
         try:
             # TODO: add database charset check later
             cnx = mysql.connector.connect(
-                host=db_info["DB_HOST"],
-                database=db_info["DB_NAME"],
-                user=db_info["DB_USER"],
-                password=db_info["DB_PASSWORD"],
+                host=self._current_effective_db_info.hostname,
```
jdkandersson, line 703: What if `self._current_effective_db_info` is `None`?

jdkandersson: I'm actually surprised `mypy` didn't complain, something to check?

tests/integration/test_core.py
```diff
@@ -131,13 +154,12 @@ async def test_openstack_object_storage_plugin(
...
         pytest.skip("no openstack configuration provided, skip openstack swift plugin setup")
-    assert ops_test.model
-    application = ops_test.model.applications[application_name]
+    application = model.applications[application_name]
     await application.set_config(
         {"wp_plugin_openstack-objectstorage_config": json.dumps(swift_config)}
     )
     # mypy has trouble to inferred types for variables that are initialized in subclasses.
-    await ops_test.model.wait_for_idle(status=ops.model.ActiveStatus.name)  # type: ignore
+    await model.wait_for_idle(status=ops.model.ActiveStatus.name)  # type: ignore
```
jdkandersson, line 162: It is tempting to define a constant somewhere like `ACTIVE_STATUS_NAME` so that we don't have to litter `type: ignore` everywhere. Thoughts? Perhaps worth a standard in the contributing guide?

tests/unit/helpers.py
```diff
@@ -0,0 +1,16 @@
...
+def get_first_endpoint(endpoints: str):
+    """Retrieve first endpoint from comma separated endpoints in host:port format.
+
+    Args:
+        endpoints: comma separated endpoints in host:port format.
+
+    Returns:
+        The first endpoint hostname.
+    """
+    return endpoints.split(",")[0].split(":")[0]
```
jdkandersson, line 16: This looks close to too dynamic for a test, should these values be hardcoded instead? Or is that not possible?

tests/unit/test_charm.py
```diff
@@ -95,50 +97,67 @@ def test_replica_consensus_stable_after_leader_reelection(
...
+    db_relation_id, db_info = setup_db_relation()
+
+    effective_db_info = charm._current_effective_db_info
+
+    assert effective_db_info.hostname == db_info["host"]
+    assert effective_db_info.database == db_info["database"]
+    assert effective_db_info.username == db_info["user"]
+    assert effective_db_info.password == db_info["password"]
+
+    harness.remove_relation(db_relation_id)
```
jdkandersson, line 125: What is the reason for needing to do this please? It looks a bit like cleanup which would be more appropriate in a fixture?

src/charm.py
yanksyoon: Yes, if any of the keys are provided, it assumes that the charm will use database connection values from configuration.

src/charm.py
yanksyoon: It's under the function `_on_relation_db_changed`, where `MySQLDatabaseChangedEvent` will store the values in stored state. I've written the reasoning inside the docstring for the `_on_relation_db_changed` function as well, which in summary is that there's no way for the charm to select which database endpoint to connect to when accessing the unit ip(set) from the databag :)

src/charm.py
yanksyoon: I believe there would be multiple endpoints if the database is running in multiple units such as cluster mode. However, I wasn't quite able to find a documentation regarding how the endpoints are supposed to be set.
What I did understand from the source code for mysql-k8s charm is that only one endpoint will be available for read-write and others would be read-only which wouldn't be appropriate for WordPress.
I've still decided to separate the endpoints by comma delimiter to cater for the definition of endpoints.

With regards to the point of having multiple read-write-able endpoints, I'm not quite sure what a good selection logic would be. Perhaps storing the already connected endpoints to the charm state and selecting the least connected endpoint?

src/charm.py
yanksyoon: This is something that I want to refactor, currently there cannot exist a case in which the `self._current_effective_db_info` can be None since it cannot progress after the `if not self._current_effective_db_info` check in `def _core_reconciliation(self)` method.

tests/integration/test_core.py
yanksyoon: That does sound pretty neat! Thank you for the suggestion!

tests/unit/test_charm.py
yanksyoon: This is needed to test the behavior after removing the relation, to verify that the charm goes back into the BlockedState, rather than to clean up :)

src/charm.py
yanksyoon: Indeed it should be!

src/charm.py
yanksyoon: Indeed we should, thank you for the suggestion :)

src/charm.py
jdkandersson: Perhaps check with the data team? 🙂

tests/unit/helpers.py
yanksyoon: Migrated the helper function to a fixture :)

src/charm.py
yanksyoon: Confirmed with the data team, [link to chat](https://chat.charmhub.io/charmhub/pl/c5ee6rkcwjr1xmehcju33gmzye) :)
Also issued a PR to improve the docstring for respective endpoints definitions.

tl;dr:
In Kubernetes charms, the endpoints is just a single service endpoint to the primary pod.

jdkandersson (approved):

weiiwang01 (approved):

metadata.yaml
yanksyoon: As per discussion, this will be another spec to discuss about the deprecation of db/mysql relation :)

PR #68, opened 2023-04-10
Fix the charm compatibility issue on focal

weiiwang01:
WordPress charm is not compatible with Python 3.8, the default Python installation on focal right now. 
Update the charm to drop support for `focal` series and change the syntax to be compatible with Python 3.8

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -6,8 +6,6 @@ bases:
     run-on:
     - name: "ubuntu"
       channel: "22.04"
-    - name: "ubuntu"
-      channel: "20.04"
 parts:
   charm:
     # Tell charmcraft to not use requirements.txt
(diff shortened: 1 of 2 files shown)


charmcraft.yaml
```diff
@@ -6,8 +6,6 @@ bases:
     run-on:
     - name: "ubuntu"
       channel: "22.04"
-    - name: "ubuntu"
```
arturo-seijas, line 9: Is this intended?

weiiwang01: This is a Kubernetes charm, I think the series is not very important. So I just remove the `20.04` to prevent this from happening again.

arturo-seijas (approved):
LGTM

nrobinaubertin (approved):

PR #70, opened 2023-04-17
Extend the database probe timeout

weiiwang01:
Currently, WordPress will retry connecting to the database for 1 minute after the database connection info is provided through the database relation. However, this 1-minute timeout is too short for certain database charms. In this commit, we extend the database probe timeout from 1 minute to 5 minutes.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -135,7 +135,7 @@ class WordpressCharm(CharmBase):
     ]
 
     _DB_CHECK_INTERVAL = 1
-    _DB_CHECK_TIMEOUT = 60
+    _DB_CHECK_TIMEOUT = 300
 
     state = StoredState()
 
--- a/tests/integration/test_error.py
+++ b/tests/integration/test_error.py
@@ -32,12 +32,12 @@ async def test_incorrect_db_config(ops_test: pytest_operator.plugin.OpsTest, app
     assert: charm should be blocked by WordPress installation errors, instead of lacking
         of database connection info.
     """
-    # Database configuration can retry for up to 60 seconds before giving up and showing an error.
+    # Database configuration can retry for up to 300 seconds before giving up and showing an error.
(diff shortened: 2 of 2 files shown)


arturo-seijas (dismissed):
LGTM

yanksyoon (dismissed):
LGTM

jdkandersson:
Looks like integration tests failed, I'm not sure why the required integration checks was skipped since it is supposed to always run, any guesses as to why?

jdkandersson:
> Looks like integration tests failed, I'm not sure why the required integration checks was skipped since it is supposed to always run, any guesses as to why?

Actually, I see the problem. The required checks here are implemented differently. @yanksyoon could we please add a ticket to the backlog to change the required status checks to be implemented in the same way as in `operator-workflows` so that we don't get the skipped problem? See here for a reference: https://github.com/canonical/operator-workflows/blob/4575dbc5fa9aea6e207d990b5b72584782812a80/.github/workflows/integration_test.yaml#L253

weiiwang01:
> Looks like integration tests failed, I'm not sure why the required integration checks was skipped since it is supposed to always run, any guesses as to why?

I think the branch protection rule `*` override the branch protection rule `main`. When was the `*` rule added?

jdkandersson:
> > Looks like integration tests failed, I'm not sure why the required integration checks was skipped since it is supposed to always run, any guesses as to why?
> 
> I think the branch protection rule `*` override the branch protection rule `main`. When was the `*` rule added?

I don't think so, GitHub applies to most specific one so the `*` should not apply. See the other comment for the reason. The check is actually marked as `required` but it was skipped which means it didn't fail so the require status checks to pass didn't work

jdkandersson:
The reason we have the `*` rule is to enforce signed commits

arturo-seijas (approved):
LGTM

yanksyoon (approved):
LGTM!

PR #71, opened 2023-04-18
fix: use edge version of loki-k8s to stabilise integration tests

yanksyoon:
The latest/stable version of loki-k8s charm isn't quite stable due to the missing `can_connect` guard (please refer to https://chat.charmhub.io/charmhub/pl/xgqyman4btym3ff74fcjomzxbw).

This PR proposes to use the latest revision (82) of the loki-k8s charm to resolve the issues with flaky tests.

Other minor improvements include:
- Updating pre-run script to remove duplicate `microk8s enable registry` (already enabled on operator-workflows)
- Updating Required Status Checks action
- Updating libraries
- Updating typing syntax to match Python 3.8
- Updating libraries

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -464,7 +464,7 @@ async def loki_fixture(
     model: Model, application_name: str
 ) -> typing.AsyncGenerator[Application, None]:
     """Deploy and return loki charm application with relation to WordPress charm."""
-    loki = await model.deploy("loki-k8s", channel="stable", trust=True)
+    loki = await model.deploy("loki-k8s", channel="edge", trust=True)
     await loki.relate(LOKI_RELATION_NAME, f"{application_name}:{LOKI_RELATION_NAME}")
     yield loki
     await loki.remove_relation(LOKI_RELATION_NAME, f"{application_name}:{LOKI_RELATION_NAME}")
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -421,18 +421,7 @@ async def build_and_deploy_fixture(
         build_and_deploy_wordpress(),
         deploy_and_wait_for_mysql_pod(),
         model.deploy("charmed-osm-mariadb-k8s", application_name="mariadb"),
-        # temporary fix for the CharmHub problem
-        ops_test.juju(
-            "deploy",
-            "nginx-ingress-integrator",
-            "ingress",
-            "--channel",
-            "edge",
-            "--series",
-            "focal",
-            "--trust",
-            check=True,
-        ),
+        model.deploy("nginx-ingress-integrator", series="focal", trust=True),
     )
     await model.wait_for_idle()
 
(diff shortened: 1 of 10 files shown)


arturo-seijas (dismissed):
LGTM

arturo-seijas (dismissed):

tests/integration/conftest.py
```diff
@@ -449,7 +440,7 @@ async def prometheus_fixture(
     model: Model, application_name: str
 ) -> typing.AsyncGenerator[Application, None]:
     """Deploy and yield prometheus charm application with relation to WordPress charm."""
-    prometheus = await model.deploy("prometheus-k8s", channel="stable", trust=True)
+    prometheus = await model.deploy("prometheus-k8s", channel="latest/edge", trust=True)
```
yhaliaw, line 443: Is this change to `latest/edge` temporary, like for Loki?
If so it might be good to add a comment, just like for Loki.

tests/integration/conftest.py
```diff
@@ -475,7 +469,7 @@ async def grafana_fixture(
     model: Model, application_name: str
 ) -> typing.AsyncGenerator[Application, None]:
     """Deploy and return grafana charm application with relation to WordPress charm."""
-    grafana = await model.deploy("grafana-k8s", channel="stable", trust=True)
+    grafana = await model.deploy("grafana-k8s", channel="latest/edge", trust=True)
```
yhaliaw, line 472: Same as above.

tests/integration/test_core.py
```diff
@@ -576,7 +577,29 @@ async def test_loki_integration(
     assert: loki joins relation successfully, logs are being output to container and to files for
         loki to scrape.
     """
-    await model.wait_for_idle(apps=[application_name, loki.name], status="active")
+
+    async def wait_loki_unit_agent_idle():
```
amandahla, line 581: I'm curious: Why is this needed?

yanksyoon: The agent can still be executing when the `model.wait_for_idle` returns (this function isn't quite 100%), and when the assertions (querying from the loki api) are executed, it returns the wrong result which doesn't contain the access logs and error logs. It's a messy workaround for having no `agent.wait_for_idle()` or `unit.wait_for_idle().
I've tested the version with and without on the actions (the tests always passes on my local ;_;), the pass rate with this workaround is much higher than the ones without. :)
Thanks for the review!

tests/integration/conftest.py
yanksyoon: Added comments for all the `edge` deployments. Thanks for the suggestion!

tests/integration/conftest.py
yanksyoon: Added comments for all the `edge` deployments. Thanks for the suggestion!

tests/integration/test_core.py
amandahla: It's a nice workaround :) What do you think about adding a small comment regarding that? Usually, we see only the model.wait_for_idle.

yanksyoon: Refactored to a helper with the description in the docstring! Thanks for the suggestion :)

weiiwang01 (approved):

nrobinaubertin (approved):

arturo-seijas (approved):

PR #77, opened 2023-05-01
Fix tutorial command for getting the pod's ip

arturo-seijas:


docs/tutorial/getting-started.md
```diff
@@ -27,10 +27,10 @@ juju add-model wordpress-tutorial
 
 ### Deploy wordpress-k8s charm
 
-Deployment of WordPress requires a database integration. The preferred method in juju is through an
-integration. In this case, a `mysql` [interface](https://juju.is/docs/sdk/integration) is required
-by the wordpress-k8s charm and hence, [`mysql-k8s`](https://charmhub.io/mysql-k8s) charm will be
-used.
+Deployment of WordPress requires a database integration. The preferred method to provide it 
+in juju is through a elation. In this case, a `mysql` [interface](https://juju.is/docs/sdk/integration)
```
mthaddon, line 31: We've switched from saying an integration here to a relation (or elation)

docs/tutorial/getting-started.md
```diff
@@ -27,10 +27,10 @@ juju add-model wordpress-tutorial
...
+is required by the wordpress-k8s charm and hence, [`mysql-k8s`](https://charmhub.io/mysql-k8s) 
+charm will be used.
```
weiiwang01, line 33: Hi, could we keep using the name "integration" when referring the juju relation?

docs/tutorial/getting-started.md
arturo-seijas: Ok, so which one of the two "integraiton" appearances should we reword, and how?

docs/tutorial/getting-started.md
arturo-seijas: Ok, so which one of the two "integraiton" appearances should we reword, and how?

weiiwang01 (approved):

mthaddon (approved):

PR #83, opened 2023-05-25
fix(ISD-413): Upgrade apache-prometheus-exporter-image

amandahla:
This PR changes the apache-prometheus-exporter-image version to a newer one with fewer vulnerabilities.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -50,7 +50,7 @@ resources:
     type: oci-image
     description: Prometheus exporter for apache
     auto-fetch: true
-    upstream-source: bitnami/apache-exporter:0.11.0
+    upstream-source: bitnami/apache-exporter:0.13.3
 
 provides:
   metrics-endpoint:


weiiwang01 (approved):

metadata.yaml
```diff
@@ -50,7 +50,7 @@ resources:
     type: oci-image
     description: Prometheus exporter for apache
     auto-fetch: true
-    upstream-source: bitnami/apache-exporter:0.11.0
+    upstream-source: bitnami/apache-exporter:0.13.3
```
jdkandersson, line 53: Similar to other PR, are there any trivy ignores we could remove now?

amandahla: Only images resulting from ROCKS or from Dockerfiles are scanned by Trivy. External ones set on upstream-source in metadata file are not scanned.

jdkandersson (approved):

PR #86, opened 2023-06-06
Ignore guzzlehttp CVE in trivy scan

arturo-seijas:
Ignore guzzlehttp CVE in trivy scan


yanksyoon (approved):
LGTM!

jdkandersson:
What is the reason please?

arturo-seijas:
> What is the reason please?

Trivy is detecting a new vulnerability in one of the wordpress' dependencies. We can't patch it, we'll need to update wordpress

gregory-schiano:
The build of the image still fails due to the snakeoil key, I think another ignore will have to be added :/

amandahla:
The config file should be considered now, can you try it, please?
See https://github.com/aquasecurity/trivy-action/issues/238

arturo-seijas:
> The config file should be considered now, can you try it, please? See [aquasecurity/trivy-action#238](https://github.com/aquasecurity/trivy-action/issues/238)

It hasn't :( I'll pin to the earlier version for now

amandahla:
My mistake, I saw the merge and then thought that it was related to the action :|

amandahla (approved):

PR #87, opened 2023-06-16
Remove the references to manual database configuration from documentation

mthaddon:
Currently our documentation shows how to connect to MySQL via a relation, but also manually.

The manual DB connection will soon be removed from the charm, but we want to remove it from the documentation first.


mthaddon:
I'm closing this out and will open a PR from a branch rather than a fork, and create a story to update our CI so that we can run the tests that don't require secrets from forks.

PR #93, opened 2023-06-28
Add database reference to integrations doc

gtrkiller:
After removing the old db relation, the integration doc needs a reference to the new database relation.


yanksyoon (dismissed):
Thanks for updating the docs!

yhaliaw (dismissed):

docs/reference/integrations.md
```diff
@@ -50,3 +50,18 @@ to the dashboard can be made but will not be persisted upon restart/redeployment
...
 Grafana-dashboard relate command: `juju relate wordpress-k8s grafana-dashboard`
+
+### database:
+
+_Interface_: mysql_client
+_Supported charms_: [Charmed MySQL](https://github.com/canonical/mysql-operator)
+
+Database endpoint can be related to mysql based charms, providing long term storage for wordpress.
+Database relation connect wordpress-k8s with charms that support the `my_sql_client` interface on port 3306
+in the database side. See [configuration](https://charmhub.io/wordpress-k8s/configure) for more detail.
```
weiiwang01, line 61: I assume the configurations mentioned here are `mysql-interface-database` and `mysql-interface-user`.
But I can't find any usage related to these two configurations in the code base, which is strange.

docs/reference/integrations.md
```diff
@@ -50,3 +50,18 @@ to the dashboard can be made but will not be persisted upon restart/redeployment
 
 Grafana-Prometheus relate command: `juju relate grafana-k8s:grafana-source prometheus-k8s:grafana-source`  
 Grafana-dashboard relate command: `juju relate wordpress-k8s grafana-dashboard`
+
+### database:
+
+_Interface_: mysql_client
+_Supported charms_: [Charmed MySQL](https://github.com/canonical/mysql-operator)
```
gregory-schiano, line 57: Shouldn't we link to charmhub instead of github ?
Also we could reference Charmed Mysql K8S

docs/reference/integrations.md
```diff
@@ -50,3 +50,18 @@ to the dashboard can be made but will not be persisted upon restart/redeployment
...
+
+Database endpoint can be related to mysql based charms, providing long term storage for wordpress.
+Database relation connect wordpress-k8s with charms that support the `my_sql_client` interface on port 3306
```
gregory-schiano, line 60: There's a typo: `mysql_client` instead of `my_sql_client`

docs/reference/integrations.md
mthaddon: Yeah, I don't think we should be referencing the configuration here either way. We should only be highlighting the new relation, which doesn't require any configuration.

docs/reference/integrations.md
```diff
@@ -50,3 +50,18 @@ to the dashboard can be made but will not be persisted upon restart/redeployment
...
+in the database side. See [configuration](https://charmhub.io/wordpress-k8s/configure) for more detail.
+
+Another way to establish database relation is to supply `db_host`, `db_name`, `db_user`, `db_password`
+configuration parameters to the charm with a MySQL database. See Configuration section for detailed
+information regarding each of the parameters.
```
mthaddon, line 65: We shouldn't document this. We'll be dropping support for these very soon.

docs/reference/integrations.md
mthaddon: I agree with both of Gregory's comments here

gregory-schiano (approved):
LGTM

mthaddon (approved):

PR #104, opened 2023-07-20
Improve the test reliability by modularization

weiiwang01:
A variety of measures to improve the reliability of integration tests include:

- Increasing the margin of safety when waiting
- Providing a more accurate timeout deadline in the charm code
- Reducing the amount of unnecessary and insignificant integration tests
- Modularizing integration tests


arturo-seijas (dismissed):
Nice! May I also suggest to make the changes specified in https://github.com/canonical/operator-workflows/pull/162?

weiiwang01:
> Nice! May I also suggest to make the changes specified in [canonical/operator-workflows#162](https://github.com/canonical/operator-workflows/pull/162)?
 
I have already added the `--charm-file` option, but the wordpress is using the `integration_test_run` workflow instead of `integration_test` workflow for extra customization. I think I remove the `db_*` configuration from the charm which will allow us to switch to the `integration_test` workflow in the following pull request.

arturo-seijas:
> > Nice! May I also suggest to make the changes specified in [canonical/operator-workflows#162](https://github.com/canonical/operator-workflows/pull/162)?
> 
> I have already added the `--charm-file` option, but the wordpress is using the `integration_test_run` workflow instead of `integration_test` workflow for extra customization. I think I remove the `db_*` configuration from the charm which will allow us to switch to the `integration_test` workflow in the following pull request.

I hadn't noticed that. That sounds like a good idea

amandahla (dismissed):

jdkandersson (dismissed):

tests/integration/test_error.py
```diff
@@ -4,28 +4,12 @@
...
-            "db_name": "test_db_name",
-            "db_user": "test_db_user",
-            "db_password": "test_db_password",
-        }
-    ],
-    indirect=True,
-    scope="function",
-)
-async def test_incorrect_db_config(ops_test: pytest_operator.plugin.OpsTest, application_name):
+@pytest.mark.usefixtures("wordpress")
```
jdkandersson, line 11: Is the `usefixtures` required given that it is being taken as an argument?

jdkandersson (commented):
Would be good for @yanksyoon so also review this PR

tests/integration/test_error.py
weiiwang01: Yes, it's unnecessary, removed, thanks!

gtrkiller (approved):
LGTM

arturo-seijas (approved):
Great!

PR #106, opened 2023-08-10
Remove DB static config

gtrkiller:
Remove the possibility to configure the DB with Juju config options and force the charm to use relation data.


tests/unit/conftest.py
```diff
@@ -144,6 +201,33 @@ def _setup_database_relation():
     return _setup_database_relation
 
 
+@pytest.fixture(scope="function")
+def setup_database_relation_connection_error(
+    harness: ops.testing.Harness, example_database_info_connection_error: typing.Dict[str, str]
+):
+    """Yields a function that can be used to set up database relation with a non 3306 port.
```
yanksyoon, line 208: I think this is better suited as `Returns a function ...` since `yield` causes the return type to be a Generator but here a function is returned. :D

yanksyoon (dismissed):
Minor nitpick :D LGTM!

jdkandersson (dismissed):
Looks good after comment from @yanksyoon is resolved

yanksyoon (approved):

jdkandersson (approved):

PR #118, opened 2023-08-30
feat(SSO): Add SSO with debug

merkata:
Putting this to get a tmate debug session and see why the SAML plugin crashes, no review required.


merkata:
Closing out, this was used for testing purposes.

PR #135, opened 2023-09-22
Fix the apache configuration test

weiiwang01:
### Overview

At present, the `_apache_config_is_enabled` function consistently returns `False`, even when the configuration is enabled. This pull request aims to rectify that issue.

### Rationale

The `_apache_config_is_enabled` function doesn't work as intended.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1211,7 +1211,7 @@ class WordpressCharm(CharmBase):
             True if certain apache config is enabled.
         """
         enabled_config = self._container().list_files("/etc/apache2/conf-enabled")
-        return f"{conf_name}.conf" in enabled_config
+        return f"{conf_name}.conf" in [file.name for file in enabled_config]
 
     def _apache_enable_config(self, conf_name: str, conf: str) -> None:
         """Create and enable an apache2 configuration file.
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -1390,11 +1394,6 @@ def _reconciliation(self, _event: EventBase) -> None:
             _event: Event fired by juju on WordPress charm related state change.
         """
         logger.info("Start reconciliation process, triggered by %s", _event)
-        if not self.state.started:
```
arturo-seijas, line 1393: You can probably remove the logic in line 150
```
self.state.set_default(
    started=False,
)
```
and in line 186
```
def _on_start(self, _event: StartEvent):
    """Record if the start event is emitted."""
    self.state.started = True
```

yanksyoon (approved):
LGTM

src/charm.py
weiiwang01: Yes, that was originally used to prevent a problem in Juju, causing applications not to complete the `start` event within the grace period to be rescheduling by Kubernetes. However, I just found out that the reconciliation process is no longer dependent on the `start` event, and I don't know when that was removed. But I think it's okay for now, and @yanksyoon will have some pull requests to remove those entirely, along with the database connectivity test and deferring for storage ready.

arturo-seijas (approved):

Issue #145, opened 2023-10-16
reconciliation loops runs three times on charm upgrade


mthaddon:
### Bug Description

The reconciliation loop runs three times on charm upgrade. See https://pastebin.canonical.com/p/3CJxW9mB4m/ for an example (sorry, Canonical internal only).

### To Reproduce

1. Deploy wordpress according to the docs
2. Trigger a charm upgrade
3. Observe the following two entries in the logs:
```
Start reconciliation process, triggered by <ConfigChangedEvent via WordpressCharm/on/config_changed[6]>
Start reconciliation process, triggered by <PebbleReadyEvent via WordpressCharm/on/apache_prometheus_exporter_pebble_ready[16]>
Start reconciliation process, triggered by <PebbleReadyEvent via WordpressCharm/on/wordpress_pebble_ready[21]>
```

### Environment

Juju 2.9.44 on OpenStack, but should be reproducible on any environment

### Relevant log output

```shell
See above
```

### Additional context

N/A

weiiwang01:
I believe this is the expected behavior for WordPress. The reconciliation process for WordPress is designed to be lightweight, checking before making any modifications. This ensures updates are applied only when necessary, so each reconciliation shouldn't consume more resources than a typical web request. This design ensures that the WordPress charm remains stateless for simplicity.

cbartz:
Closed due explanatory comment https://github.com/canonical/wordpress-k8s-operator/issues/145#issuecomment-1765722052 .

PR #154, opened 2023-10-25
Change uploads directory ownership

amandahla:
Applicable spec: <link>

### Overview

The uploads directory is mounted with permission only to the root user. This prevents the correct behavior of uploading files via WordPress. This PR fixes that by changing the ownership using MongoDB charm as [reference](https://github.com/canonical/mongodb-k8s-operator/blob/main/src/charm.py#L1089).

### Rationale

Allow uploads directory to be writable to Wordpress user as expected.

### Juju Events Changes

N/A

### Module Changes

N/A

### Library Changes

N/A

### Checklist

- [X] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [X] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [X] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [X] The documentation is generated using `src-docs`
- [X] The documentation for charmhub is updated.
- [X] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)


arturo-seijas (commented):
Wouldn't it be simpler to do this on the rock?

amandahla:
> Wouldn't it be simpler to do this on the rock?

The uploads is a mounted storage, so doing in the ROCK would only affect the directory before mounting the storage.

arturo-seijas (dismissed):
LGTM

gregory-schiano (commented):
Should we add some tests ?

weiiwang01 (approved):

weiiwang01 (dismissed):

weiiwang01:
I am a little bit curious why this was not a problem before, anyway the pull request lgtm, thanks!

weiiwang01:
> I am a little bit curious why this was not a problem before, anyway the pull request lgtm, thanks!

Did a little digging, and it seems like the `hostpath-provisioner` in microk8s is running with a umask of `0000` 

```
$ cat /proc/$(pidof /hostpath-provisioner)/status | grep -i umask
Umask:  0000
```

so the directory created for `hostpath` storage will have a permission of `777`

```
$ ls -lah /var/snap/microk8s/common/default-storage/
total 140K
drwxrwxrwx  2 root   root   4.0K Oct 26 16:22 test-wordpress-wordpress-k8s-uploads-31e96c3e-wordpress-k8s-0-pvc-3b6a4639-3308-49c8-9284-ee9c05a702be
```

amandahla:
> > I am a little bit curious why this was not a problem before, anyway the pull request lgtm, thanks!
> 
> Did a little digging, and it seems like the `hostpath-provisioner` in microk8s is running with a umask of `0000`
> 
> ```
> $ cat /proc/$(pidof /hostpath-provisioner)/status | grep -i umask
> Umask:  0000
> ```
> 
> so the directory created for `hostpath` storage will have a permission of `777`
> 
> ```
> $ ls -lah /var/snap/microk8s/common/default-storage/
> total 140K
> drwxrwxrwx  2 root   root   4.0K Oct 26 16:22 test-wordpress-wordpress-k8s-uploads-31e96c3e-wordpress-k8s-0-pvc-3b6a4639-3308-49c8-9284-ee9c05a702be
> ```

Wow, great finding!

weiiwang01 (dismissed):

src/charm.py
```diff
@@ -1406,6 +1407,12 @@ def _reconciliation(self, _event: EventBase) -> None:
             self.unit.status = WaitingStatus("Waiting for storage")
             _event.defer()
             return
+        # Ensure the data directory for WordPress is writable for the WordPress user.
+        self._container().exec(
+            f"chown {self._WORDPRESS_USER}:{self._WORDPRESS_GROUP} -R {self._WP_UPLOADS_PATH}".split(
```
weiiwang01, line 1412: I think we might just need to change the ownership of the mounting directory, i.e. without `-R`. Since recursively change the ownership of all files can be slow if there are a lot files in the directory.

src/charm.py
```diff
@@ -1406,6 +1407,12 @@ def _reconciliation(self, _event: EventBase) -> None:
...
+                " "
```
weiiwang01, line 1413: nit: calling `.split()` without an argument is similar to `.split(" ")`, or we can just provide a list to `exec` instead of splitting the argument string as the argument list is not very long.

src/charm.py
amandahla: Its definitely a good point but since the _WORDPRESS_USER was changed recently from www-data to _daemon_, I was afraid we might have a case where files would be with the previous permission. What do you think?

amandahla: BTW there is no wait for output, do you think it will affect the charm anyway?

src/charm.py
amandahla: Nice, changed.

arturo-seijas (dismissed):
LGTM

yanksyoon (dismissed):
LGTM, thanks for the changes!

src/charm.py
weiiwang01: I think the `wait` might be necessary, `exec` without `wait` might be undefined behavior.

Yes, you are correct the upgrade might cause some problem. Could we run the recursive version only on upgrade event so we can keep the reconciliation process as light as possible as the reconciliation can run multiple times during the charm's lifecycles https://github.com/canonical/wordpress-k8s-operator/issues/145

amandahla: I see. I made some changes now, what do you think?

weiiwang01: 👍

weiiwang01 (approved):

src/charm.py
```diff
@@ -1406,6 +1436,7 @@ def _reconciliation(self, _event: EventBase) -> None:
             self.unit.status = WaitingStatus("Waiting for storage")
             _event.defer()
             return
+        self._change_uploads_directory_ownership()
```
yanksyoon, line 1439: I'm wondering if changing uploads directory ownership needs to run every reconcile, can be discussed for later issues though!

yanksyoon (approved):
Minor comment on reconcile loop, LGTM!

src/charm.py
amandahla: Agree, maybe this could be discussed in the story ISD-721 (refactor). I added a link to this comment there, thanks!

Issue #163, opened 2023-11-13
No more relational databag usage


juditnovak:
### Bug Description

Currently the wordpress(-k8s) charm is reading information from the databag.
This results in the charm hanging in the `_test_database_connectivity()` function at reconciliation, when the service would start up.

https://github.com/canonical/wordpress-k8s-operator/blob/main/src/charm.py#L612-L613

This and similar references direct references to the databag must be changed to use the `fetch_relation_data()` (or the new, more convenient `fetch_relation_field()` interface function call.

Please feel free to reach out if more info or help may be needed.

### To Reproduce

juju deploy wordpress-k8s --channel latest/edge

juju deploy mysql-router-k8s --trust --channel 8.0/edge
juju deploy mysql-k8s --trust --channel 8.0/edge

juju relate mysql-router-k8s mysql-k8s
juju relate mysql-router-k8s wordpress-k8s

### Environment

Juju 3.1

### Relevant log output

```shell
Typically the "last words" wordpress may say before hanging:

unit-wordpress-k8s-0: 23:29:56 INFO unit.wordpress-k8s/0.juju-log database:7: Ensure WordPress (apache) server is down
unit-wordpress-k8s-0: 23:30:05 INFO unit.wordpress-k8s/0.juju-log database:7: Update wp-config.php content in container
unit-wordpress-k8s-0: 23:30:17 INFO unit.wordpress-k8s/0.juju-log database:7: Ensure WordPress server is up
```

juditnovak:
The issue was resolved in https://github.com/canonical/wordpress-k8s-operator/pull/166

PR #166, opened 2023-11-14
[DPE-3009] No more relational databag usage

juditnovak:
Overview
This PR is addressing issue https://github.com/canonical/wordpress-k8s-operator/issues/163

Rationale
The Relational Databag is not to be addressed directly but only via interface functions.
Part of the relational data is now "hiding" in Juju Secrets and are not exposed on the databag anymore.
The data_platform_libs/data_interfaces DatabaseRequires and DatabaseProvies interfaces ensure transparent and safe access to all relational data. Outside of these functions, no access/manipulation of this data can be guaranteed in the future.

Module Changes
Relational Databag usage is replaced with DatabaseRequires interface calls.

### DIFF
--- a/.github/workflows/integration_test_juju3.yaml
+++ b/.github/workflows/integration_test_juju3.yaml
@@ -22,6 +22,7 @@ jobs:
           sudo microk8s kubectl -n kube-system rollout status -w deployment/hostpath-provisioner
           sudo microk8s config > ${GITHUB_WORKSPACE}/kube-config
           chmod +x tests/integration/pre_run_script_juju3.sh
-          ./tests/integration/pre_run_script_juju3.sh"
+          ./tests/integration/pre_run_script_juju3.sh
+          sed -i 's/juju>=2.9,<3/juju>=3.2.2/' ./tests/integration/requirements.txt"
       setup-devstack-swift: true
       trivy-image-config: ./trivy.yaml
--- a/.github/workflows/integration_test_juju3.yaml
+++ b/.github/workflows/integration_test_juju3.yaml
@@ -21,7 +21,7 @@ jobs:
         -c "sudo microk8s enable hostpath-storage
           sudo microk8s kubectl -n kube-system rollout status -w deployment/hostpath-provisioner
           sudo microk8s config > ${GITHUB_WORKSPACE}/kube-config
-          chmod +x tests/integration/pre_run_script.sh
+          chmod +x tests/integration/pre_run_script_juju3.sh
           ./tests/integration/pre_run_script_juju3.sh"
       setup-devstack-swift: true
       trivy-image-config: ./trivy.yaml
--- a/.github/workflows/integration_test_juju3.yaml
+++ b/.github/workflows/integration_test_juju3.yaml
@@ -1,4 +1,4 @@
-name: integration-tests
+name: integration-tests-juju3
 
 on:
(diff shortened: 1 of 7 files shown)


arturo-seijas (dismissed):
LGTM. Thanks for your contribution

.github/workflows/integration_test_juju3.yaml
```diff
@@ -0,0 +1,26 @@
...
+  integration-test-juju3:
+    name: Integration Test on Juju3
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
+    secrets: inherit
+    with:
+      extra-arguments: >-
+        --openstack-rc=${GITHUB_WORKSPACE}/openrc
+        --kube-config=${GITHUB_WORKSPACE}/kube-config
+        --screenshot-dir=/tmp
+      juju-channel: 3.1/stable
```
arturo-seijas, line 17: You'll probably  need to specify the microk8s channel too `channel: 1.28-strict/stable`

.github/workflows/integration_test_juju3.yaml
```diff
@@ -0,0 +1,28 @@
...
+      channel: 1.27-strict/stable
+      modules: '["test_addon", "test_core"]'
+      pre-run-script: |
+        -c "sudo microk8s enable hostpath-storage
+          sudo microk8s kubectl -n kube-system rollout status -w deployment/hostpath-provisioner
+          sudo microk8s config > ${GITHUB_WORKSPACE}/kube-config
+          chmod +x tests/integration/pre_run_script_juju3.sh
+          ./tests/integration/pre_run_script_juju3.sh
+          sed -i 's/juju>=2.9,<3/juju>=3.2.2/' ./tests/integration/requirements.txt"
```
weiiwang01, line 26: I think using `sed` to modify the version number in `requirements.txt` could be not very reliable due to potential changes in the Juju requirement declaration. I think it would be better to create a distinct `requirements-juju3.txt` specifically for Juju 3 and create a new tox environment, such as `integration-juju3` which would use the newly created `requirements-juju3.txt`. Additionally, the `test-tox-env` can be used to switch the workflow to use the new testing tox environment as needed.

juditnovak: Sure, no worries.
I was aiming for a temporary solution here (assuming U Guys may fully switch to Juju3 soon).
However if Juju2 is to be supported in a long term, it's indeed better to have it reflected up to `tox` level

src/charm.py
```diff
@@ -299,24 +295,52 @@ def _wp_generate_password(length: int = 64) -> str:
...
+        sensitive data -- may come from Juju secrets. To be accessed ONLY via
+        DatabaseRequires public 'fetch' functions.
+        Used by WordPress to enhance security and must be synchronized.
+
+        Returns:
+            'dict' snapshot of data shared among WordPress application replicas.
+        """
+        return self.database.fetch_relation_data([self.replica_relation.id]).get(
+            self.replica_relation.id
+        )
```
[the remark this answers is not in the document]
juditnovak: You are TOTALLY right. I apologize, I didn't pay sufficent attention on the context. Fixed now

weiiwang01 (approved):

merkata (approved):
Thank you for the contribution, lgtm!

PR #172, opened 2023-11-23
Fix default theme and plugin list inconsistency

weiiwang01:
### Overview

WordPress charm defines a `_WORDPRESS_DEFAULT_THEMES` and `_WORDPRESS_DEFAULT_PLUGINS` list containing all default installed themes and plugins in the workload image. These lists serve as the reference for theme and plugin management so it must alignment with themes and plugins within the OCI image.

Update the two list to resolve problems caused by the inconsistency.

**Reference for obtaining the lists of default themes and plugins:**

```bash
# build the WordPress image
rockcraft pack
skopeo --insecure-policy copy oci-archive:$(ls *.rock) docker-daemon:wordpress-image:latest

# start a mysql database for WordPress
docker run -d -e MYSQL_USER=wordpress -e MYSQL_PASSWORD=wordpress -e MYSQL_DATABASE=wordpress -e MYSQL_ROOT_PASSWORD=wordpress --name=mysql mysql:latest

# install WordPress and print all themes and plugins
docker run -it --rm --entrypoint /bin/bash --workdir /var/www/html wordpress-image:latest
wp config create --dbname=wordpress --dbuser=wordpress --dbpass=wordpress --dbhost=<mysql-ip>
wp core install --url=example.com --title=test --admin_user=admin --admin_email=admin@example.com
wp theme list
wp plugin list

# clean up
docker rm -f mysql
```

### Checklist

### DIFF
--- a/wordpress_rock/rockcraft.yaml
+++ b/wordpress_rock/rockcraft.yaml
@@ -81,8 +81,6 @@ parts:
       mkdir -p wordpress_install_dir
       (cd wordpress_install_dir; $CRAFT_PART_BUILD/wp core download --version=${WP_VERSION} --allow-root)
 
-      rm -rf wordpress_install_dir/wp-content/themes/twentytwentythree
-      rm -rf wordpress_install_dir/wp-content/themes/twentytwentyfour
       cp -R . $CRAFT_PART_INSTALL
     organize:
       wordpress_install_dir: /var/www/html
--- a/src/charm.py
+++ b/src/charm.py
@@ -76,10 +76,9 @@ class WordpressCharm(CharmBase):
         "resource-centre",
         "thematic",
         "twentyeleven",
-        "twentytwenty",
-        "twentytwentyone",
-        "twentytwentytwo",
+        "twentytwentyfour",
         "twentytwentythree",
+        "twentytwentytwo",
         "ubuntu-cloud-website",
         "ubuntu-community-wordpress-theme/ubuntu-community",
         "ubuntu-community/ubuntu-community",
@@ -103,6 +102,7 @@ class WordpressCharm(CharmBase):
         "favicon-by-realfavicongenerator",
         "feedwordpress",
         "genesis-columns-advanced",
+        "hello",
         "line-break-shortcode",
         "wp-mastodon-share",
         "no-category-base-wpml",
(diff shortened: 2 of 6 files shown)


arturo-seijas (dismissed):

varshigupta12 (dismissed):

wordpress_rock/rockcraft.yaml
```diff
@@ -81,8 +81,6 @@ parts:
       mkdir -p wordpress_install_dir
       (cd wordpress_install_dir; $CRAFT_PART_BUILD/wp core download --version=${WP_VERSION} --allow-root)
 
-      rm -rf wordpress_install_dir/wp-content/themes/twentytwentythree
-      rm -rf wordpress_install_dir/wp-content/themes/twentytwentyfour
```
Thanhphan1147, line 85: Hi, I think there a a few problems with these 2 themes specifically. the `twentytwentythree` theme that comes by default with wordpress seems to be faulty (when you first go to the home page it just shows a blank screen, you'd need to navigate to /wp-admin.php to see the theme automatically uninstalls itself before the homepage can work again).

For the `twentytwentyfour` theme there seems to be a bug in our integration test (specifically this [line](https://github.com/canonical/wordpress-k8s-operator/blob/ab1dbde93509c4a5cd95d4cc65f2b31e399ba415/tests/integration/helper.py#L113) does not work well with the new theme - maybe the API changed?)

Thanhphan1147: These 2 lines are workarounds for those problems, by removing `twentytwentythree` and lets wordpress install it during reconciliation. 

for `twentytwentyfour` we'd need to fix the integration test before adding it back in

weiiwang01: > theme automatically uninstalls

I think this is caused by the theme not in the `_DEFAULT_WORDPRESS_THEMES` list, because the WordPress charm will remove all theme that's not in the `_DEFAULT_WORDPRESS_THEMES` list during the reconciliation, and if the WordPress active theme not in the list, this will cause the WordPress page to go blank.

weiiwang01: > for twentytwentyfour we'd need to fix the integration test before adding it back in

Yes, I think I have found the reason for the failing, the `twentytwentyfour` theme came with WordPress 6.4 has a redesign of the home page, I have updated the integration test to accommodate that.

weiiwang01: All the problems in the integration tests have been fixed, I have removed the `test_default_wordpress_themes_and_plugins` test because that test only sees the themes and plugins after the reconciliation which is kind pointless. Sorry I missed your original pull request.

Thanhphan1147 (approved):
LGTM, thanks!

mthaddon (approved):

PR #174, opened 2023-11-27
Mandate host and port in database information

weiiwang01:
### Overview

User feedback indicates that the WordPress charm might initiate the reconciliation process without a database endpoint provided in the database relation, leading to errors during database connectivity verification. Although we haven't been able to replicate this issue, updating the charm to prevent it from proceeding to reconciliation in the absence of a database endpoint in the database relation, as a precautionary measure.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -603,6 +603,10 @@ class WordpressCharm(CharmBase):
             self.database.fetch_relation_field(relation.id, "endpoints")
         )
         if host is None:
+            logger.warning(
+                "`endpoints` is missing in database relation data: %s",
+                dict(relation.data[relation.app]),
+            )
             return None
         return types_.DatabaseConfig(
             hostname=host,
--- a/src/charm.py
+++ b/src/charm.py
@@ -608,6 +608,8 @@ class WordpressCharm(CharmBase):
         host, port = self._parse_database_endpoints(
             self.database.fetch_relation_field(relation.id, "endpoints")
         )
+        if host is None:
+            return None
         return types_.DatabaseConfig(
             hostname=host,
             port=port,
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -879,7 +879,7 @@ def test_wordpress_version_set(harness: ops.testing.Harness):
 
 
 @pytest.mark.usefixtures("attach_storage")
-def test_leader_installation_failure(
+def test_waiting_for_leader_installation_timeout(
     patch: WordpressPatch, harness: ops.testing.Harness, app_name
 ):
     """
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -6,6 +6,7 @@
 # pylint:disable=protected-access
 
 import json
+import secrets
 import typing
 import unittest.mock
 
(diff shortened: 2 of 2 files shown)


yanksyoon (dismissed):

jdkandersson (dismissed):

tests/unit/test_charm.py
```diff
@@ -876,3 +876,43 @@ def test_wordpress_version_set(harness: ops.testing.Harness):
...
+
+@pytest.mark.usefixtures("attach_storage")
+def test_leader_installation_failure(
+    patch: WordpressPatch, harness: ops.testing.Harness, app_name
+):
+    """
+    arrange: charm peer and database relation is ready, the storage is attached.
+    act: start the charm as a follower unit.
+    assert: charm unit should enter blocked state, and the installation error should be seen
+        in the status.
```
mthaddon, line 889: This isn't describing (to me at least) what the test is actually doing. Isn't the failure here because "endpoints" doesn't include a `:` and so `host` and `port` are set to `None`? All we're saying in `arrange` is "charm peer and database relation is ready, the storage is attached."

The name is `test_leader_installation_failure` but then we say here "start the charm as a follower unit" so that seems a bit inconsistent.

src/charm.py
```diff
@@ -602,6 +602,8 @@ def _current_effective_db_info(self) -> Optional[types_.DatabaseConfig]:
         host, port = self._parse_database_endpoints(
             self.database.fetch_relation_field(relation.id, "endpoints")
         )
+        if host is None:
+            return None
```
amandahla, line 606: What about adding some log indicating that the relation is there but the endpoint (for some reason?) is not available?

weiiwang01: Yes, indeed, updated! But I suspect anyone will look the log if there's no errors 😂

mthaddon: You'd be surprised. Quite often when things don't happen as expected even if there's no error you end up looking in the logs. Having logging like this is definitely very useful.

amandahla (dismissed):
LGTM

tests/unit/test_charm.py
```diff
@@ -876,3 +876,43 @@ def test_wordpress_version_set(harness: ops.testing.Harness):
...
+        key_values={
+            "endpoints": "test",
+            "database": "test",
+            "username": "test",
+            "password": "test",
+        },
+    )
+    # ignore bandit B106:hardcoded_password_funcarg
+    patch.database.prepare_database(
+        host="test", database="test", user="test", password="test"
```
jdkandersson, line 911: Why not use the `secrets` library to generate a random string for the password? That way we don't need a disable for a security issue

weiiwang01: I think the problem of using `secrets` library is that it's a little bit perplexing for people not knowing the intent. I once received a comment asking about why using `secrets.token_urlsafe` in tests.

jdkandersson: The explanation can be that it is a good habit not to hardcode passwords. Whilst the harm done in tests is not that bad, people could still copy paste code from tests or see it and think it is reasonable. It is also good to keep the security warning disables to a minimum so we only do it rarely and treat warnings seriously. You can always include a comment explaining why we use secrets to generate the password

weiiwang01: Gotcha, updated, thanks!

jdkandersson (dismissed):

amandahla (dismissed):
LGTM

jdkandersson (approved):

amandahla (approved):
LGTM

PR #194, opened 2024-02-28
Improve WordPress's observability and increase health check timeout

weiiwang01:
Improve WordPress's observability by updating the Apache access log format to a structured format, including the sub-millisecond request times and durations in the access log. Introduce a new metric, `apache_access_log_request_duration_microseconds`, a histogram tracking request durations. Unlike metrics available through Apache's `mod_status`, this specific metric is derived from Promtail reading the access log, which means it will only be available when `logging` integration is established. This approach should offer better performance compared to using `logql`.

Update the dashboard with a new panel to visualizes request duration percentiles, providing better insights into WordPress performance issues.

Some other miscellaneous changes include adjusting the WordPress health check timeout from 3 to 5 seconds and upgrading WordPress to patch version `6.4.3`.

Screenshot 2024-02-28 at 23 51 27

### DIFF
--- a/wordpress_rock/files/etc/apache2/sites-available/000-default.conf
+++ b/wordpress_rock/files/etc/apache2/sites-available/000-default.conf
@@ -17,10 +17,6 @@
 	# modules, e.g.
 	#LogLevel info ssl:warn
 
-	# 2023/02/14
-	# Write logs to both files for loki log scraping and stdout for kubernetes
-	# container logs. Required for loki integration.
-	ErrorLog "|$ tee {APACHE_LOG_DIR}/error.log"
 	CustomLog "${APACHE_LOG_DIR}/access.log" logfmt
 	CustomLog /dev/stdout logfmt
 
--- a/wordpress_rock/files/etc/apache2/sites-available/000-default.conf
+++ b/wordpress_rock/files/etc/apache2/sites-available/000-default.conf
@@ -17,7 +17,11 @@
 	# modules, e.g.
 	#LogLevel info ssl:warn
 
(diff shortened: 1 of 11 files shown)


arturo-seijas (dismissed):

wordpress_rock/files/etc/apache2/sites-available/000-default.conf
```diff
@@ -17,12 +17,8 @@
 	# modules, e.g.
 	#LogLevel info ssl:warn
 
-	# 2023/02/14
-	# Write logs to both files for loki log scraping and stdout for kubernetes
-	# container logs. Required for loki integration.
-	ErrorLog "|$ tee {APACHE_LOG_DIR}/error.log"
-	CustomLog "${APACHE_LOG_DIR}/access.log" combined
-	CustomLog /dev/stdout combined
+	CustomLog "| /usr/bin/rotatelogs -p /usr/bin/pruneaccesslogs ${APACHE_LOG_DIR}/access.%Y-%m-%d-%H-%M-%S.log 256M" logfmt
```
mthaddon, line 20: We seem to be missing a `tee` here - is that deliberate? Might be good to include a comment explaining what we're doing here - we seem to have removed two lines of comments above.

weiiwang01: So, for the `CustomLog`, there is a statement below (`CustomLog /dev/stdout logfmt`) that will output the access log to stdout. And, for `ErrorLog`, using `rotatelogs` with the `-e` option will function like `tee` which also outputs the log to stdout. Added some comments to explain the logging statements here. Thanks!

mthaddon (approved):

arturo-seijas (approved):

PR #200, opened 2024-03-15
Fix storage mount ownership update process during charm upgrades

weiiwang01:
Currently, when updating the storage mount ownership during the `upgrade-charm` event, the WordPress charm does not account for situations where pebble is not ready. This means that if the `upgrade-charm` event is fired before `wordpress-pebble-ready`, it will cause an error. To avoid this problem, update the process for updating storage ownership.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -368,7 +368,6 @@ class WordpressCharm(CharmBase):
             _event: required by ops framework, not used.
         """
         self._setup_replica_data(_event)
-        self._change_uploads_directory_ownership(recursive=True)
 
     def _gen_wp_config(self):
         """Generate the wp-config.php file WordPress needs based on charm config and relations.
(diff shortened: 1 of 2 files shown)


src/charm.py
jdkandersson, line 1: Should there be an integration test that ensures there is no regression in the future?

weiiwang01: There's already an integration test for this feature: `test_uploads_owner`

jdkandersson: do you think it is worth/ possible to update that test so that it would fail without this fix? Or would that be tricky to orchestrate?

weiiwang01: Yes, this depends on the order of events, we used to have an integration test for testing the upgrade charm scenario, but I don't remember that caught this problem before.

jdkandersson (approved):

arturo-seijas (approved):

PR #208, opened 2024-04-11
fix: tell type checkers that the config options are strings (trivial)

tonyandrewmeyer:
Applicable spec: N/A

### Overview

ops<=2.12 wrongly has `self.config[x]` typed as `str`, when actually it could be an int, float, bool, or str, depending on the config type. We're fixing this in [ops:#1183](https://github.com/canonical/operator/pull/1183), but that will break static checking that currently assumes that the config is a str (because ops doesn't validate the schema, so all options will be bool|int|float|str).

This PR adds a `typing.cast` call where the config values are loaded and used where the type should be str.

### Rationale

Covered in the overview.

### Juju Events Changes

N/A

### Module Changes

N/A

### Library Changes

N/A

### Checklist

- [ ] The documentation for charmhub is updated.

No docs changes required.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -958,7 +958,9 @@ class WordpressCharm(CharmBase):
         current_installed_addons = set(t["name"] for t in exec_result.result)
         logger.debug("Currently installed %s %s", addon_type, current_installed_addons)
         addons_in_config = [
-            t.strip() for t in self.model.config[f"{addon_type}s"].split(",") if t.strip()
+            t.strip()
+            for t in cast(str, self.model.config[f"{addon_type}s"]).split(",")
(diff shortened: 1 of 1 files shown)


yanksyoon (approved):

yanksyoon:
/canonical/self-hosted-runners/run-workflows 22207de61a8e8a4247506ace5d772a24d1d05ac8

yanksyoon (approved):

varshigupta12 (approved):

arturo-seijas:
/canonical/self-hosted-runners/run-workflows 2bc4cd725aa1fb273f423433e09cee339738b3ec

PR #216, opened 2024-05-08
Refactor to remove wp eval code

javierdelapuente:
Applicable spec: <link>

### Overview

Small refactor to remove `wp eval` that runs directly code against the wordpress installation to configure the `wordpress-teams-integration` plugin. 

Instead, to configure options  in the same way in all plugins, use `wp option update` command instead of `eval`. That will make it easier to have generic code, and stop using hardcoded php code to configure options in the plugin.

A change was also needed  in the plugin (already merged):
https://git.launchpad.net/~canonical-sysadmins/wordpress-teams-integration/+git/wordpress-teams-integration/commit/?id=a643325df2e40ad5a34346f93872c47cdaa09051

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1178,19 +1178,8 @@ class WordpressCharm(CharmBase):
                 f"Unable to config akismet plugin, {result.message}"
             )
 
-    def _wp_eval(self, php_code: str):
-        """Execute arbitrary PHP code.
-
-        Args:
-            php_code: PHP code to be executed.
-
-        Returns:
-            An instance of :attr:`types_.ExecResult`.
-        """
-        return self._wrapped_run_wp_cli(["wp", "eval", php_code])
-
     @staticmethod
-    def _encode_openid_team_map(team_map: str) -> str:
+    def _encode_openid_team_map(team_map: str) -> dict:
         """Convert wp_plugin_openid_team_map setting to openid_teams_trust_list WordPress option.
 
         example input: site-sysadmins=administrator,site-editors=editor,site-executives=editor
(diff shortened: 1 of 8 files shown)


arturo-seijas (approved):

src/charm.py
```diff
@@ -1199,21 +1188,20 @@ def _encode_openid_team_map(team_map: str) -> str:
...
-                f"'team'=>'{launchpad_role}',"
-                f"'role'=>'{wordpress_role}',"
-                f"'server' => '0',),"
-            )
-        return f"array({''.join(array_items)})"
+            teams_parsed[str(idx)] = {
+                "id": idx,
+                "team": launchpad_role,
+                "role": wordpress_role,
+                "server": "0",
```
yanksyoon, line 1202: Out of curiosity, what does the server represent? - (Optionally, would it be possible to have this in a typed dict with attributes so these can be explained within docstrings?)

yanksyoon (approved):
LGTM!

amandahla (approved):
LGTM

src/charm.py
javierdelapuente: Looks like it is a way of mapping servers to [open id endpoints](https://git.launchpad.net/~canonical-sysadmins/wordpress-teams-integration/+git/wordpress-teams-integration/tree/openid-teams.php?id=a643325df2e40ad5a34346f93872c47cdaa09051#n50) and 0 is the only value used... To be honest, it looks like something not used...

I will merge as I do not plan to work more in Wordpress charm, but I believe the correct action, beside documenting as you state, would be to work on the plugins...

PR #235, opened 2024-08-01
Create a charm action to run wp core update db after a wordpress upgrade

### LINKED ISSUE #233: Create a charm action to run `wp core update-db` after a WordPress upgrade
### Enhancement Proposal

After upgrading WordPress to a new version, it is typically necessary to run `wp core update-db` to migrate the database schema. Currently, there is no way for users of the WordPress charm to run this command after upgrading the charm other than using `juju ssh`.

Add a charm action to execute this command, allowing users to manually invoke the action to migrate the database after performing a juju refresh. This action should probably include an option to dry-run the command.

Alternatively, consider automatically executing this command in charm on juju refresh.

alithethird:
Added an action for updating database.
NEED ADVICE: I did not check the boxes regarding documentation because I am not sure if I need to update the documentation and how to do that.

### Overview

After updating the WordPress version we have to update the database and there was no action for it. This PR creates an action for that called 'update-database'.

### Rationale

After updating the WordPress version we have to update the database and there was no action for it. 
User had to ssh into the container to do that.

### Checklist

- [ ] The documentation for charmhub is updated.


tests/unit/test_charm.py
cbartz, line 1: Could you add a test for the non-happy path?

actions.yaml
```diff
@@ -11,3 +11,11 @@ rotate-wordpress-secrets:
...
     secure_auth_salt.
     Users will be forced to log in again. This might be useful under security breach circumstances.
+update-database:
+  description: >
+    After upgrading WordPress to a new version it is typically necessary to run 'wp core update-db'
+    to migrate the database schema. This action does exactly that.
+  params:
+    dry-run:
+      type: boolean
+      description: Runs the 'wp core update-db --dry-run' command.
```
cbartz, line 21: nit: final new line missing

tests/unit/wordpress_mock.py
```diff
@@ -645,6 +645,16 @@ def _mock_chown_uploads(self, _cmd):
         """Simulate ``chown`` command execution in the container."""
         return ExecProcessMock(return_code=0, stdout="", stderr="")
 
+    @_exec_handler.register(lambda cmd: cmd[:3] == ["wp", "core", "update-db"])
+    def _mock_wp_update_database(self, _cmd):
+        """Simulate ``wp core update-db`` command execution in the container."""
+        return ExecProcessMock(return_code=0, stdout="", stderr="")
+
+    @_exec_handler.register(lambda cmd: cmd[:3] == ["wp", "core", "update-db", "--dry-run"])
```
cbartz, line 653: Should it not be `cmd[:4]` ?

cbartz (commented):
Thank you! Some comments.

tests/unit/wordpress_mock.py
alithethird: Oh yes, it is. Missed it. Thanks!

actions.yaml
alithethird: It doesn't have a final new line in the original version. Also should all yaml files have an empty last line?

tests/unit/test_charm.py
alithethird: How do i use the `_mock_wp_update_database_fail` mock in the non-happy path test function?

alithethird: Also I don't really know in what cases that would fail :sweat_smile:

actions.yaml
cbartz: Its like an unwritten convention in our team (afaik). But if it was not there before, you can ignore my comment and close this conversation.

tests/unit/test_charm.py
alithethird: Looks like it would fail when database relation is not created.

cbartz: Probably @weiiwang01 can help you better here, but I guess your options would be to either use the patch fixture and modify the `container` or patch `WordpressCharm._run_wp_cli` to return a corresponding negative result. At first glance, the latter approach seems more feasible.

amandahla:
Wouldnt that be the case of adding this to the _on_upgrade_charm event? When running with --dry-run wouldn't be possible to check if is needed or not?

src/charm.py
```diff
@@ -368,6 +411,7 @@ def _on_upgrade_charm(self, _event: UpgradeCharmEvent):
             _event: required by ops framework, not used.
         """
         self._setup_replica_data(_event)
+        self._update_database()
```
javierdelapuente, line 414: What if this fails?

alithethird: Is logging with ERROR  a good idea for this case?

src/charm.py
```diff
@@ -267,6 +268,48 @@ def _on_rotate_wordpress_secrets_action(self, event: ActionEvent):
         self._reconciliation(event)
         event.set_results({"result": "ok"})
 
+    def _update_database(self, dry_run: bool = False) -> types_.ExecResult:
```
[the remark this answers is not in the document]
javierdelapuente: I see that it is done that way in other places... I think it would be cleaner not to return ExecResult, so as you prefer.

src/charm.py
javierdelapuente: Not sure if this is a reason to go into error state. I think it is a bit tricky...

actions.yaml
yanksyoon: Please do add a new-line :D Reason: https://gist.github.com/OleksiyRudenko/d51388345ea55767b7672307fe35adf3

docs/how-to/update-wordpress-charm.md
yanksyoon, line 1: @erinecon Hello Erin! Would you be able to take a look at the documentation update please?

src/charm.py
yanksyoon: Update db fail would be catastrophic, error would auto-retry which is not what we want... @jdkandersson what do we want to do in such cases?

yanksyoon (commented):
I think we should discuss what to do when db-update fails, otherwise LGTM!

src/charm.py
alithethird: David said we should not put this in the `on-upgrade` event. I have removed that in the latest commit

src/charm.py
```diff
@@ -268,6 +268,33 @@ def _on_rotate_wordpress_secrets_action(self, event: ActionEvent):
...
```
cbartz, line 271: Nit: Can you reorder the function location, it should be below _on_update_database_action, see also https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#function-and-method-ordering

src/charm.py
cbartz: As discussed in the APAC Stand-up, you should probably remove this as it's out of scope for this task.

tests/unit/test_charm.py
```diff
@@ -487,31 +488,33 @@ def test_update_database(
...
+def test_update_database_fail(
+    patch,
     harness: ops.testing.Harness,
     action_event_mock: unittest.mock.MagicMock,
 ):
     """
-    arrange: after charm is initialized and database ready.
-    act: run update-database action with dry-run.
-    assert: update-database action should success and return "ok".
+    arrange: after charm is initialized and database is not ready.
```
cbartz, line 505: Perhaps more explicitly: "and the database update is mocked to fail".

tests/unit/wordpress_mock.py
```diff
@@ -645,6 +645,13 @@ def _mock_chown_uploads(self, _cmd):
         """Simulate ``chown`` command execution in the container."""
         return ExecProcessMock(return_code=0, stdout="", stderr="")
 
+    @_exec_handler.register(lambda cmd: cmd[:3] == ["wp", "core", "update-db"])
+    def _mock_wp_update_database(self, _cmd):
+        """Simulate ``wp core update-db`` command execution in the container."""
+        if self._fail_wp_update_database:
```
cbartz, line 651: What happens if `_fail_wp_update_database_` is not initialised? You should probably initialise it in the constructor.

src/charm.py
alithethird: Thanks! Done.

tests/unit/test_charm.py
alithethird: Thanks! Done.

tests/unit/wordpress_mock.py
alithethird: Done! Thanks!!

src/charm.py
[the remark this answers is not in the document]
alithethird: Updated the private function and action function to make them similar to how other actions are done in the charm but they used ExecResult's in other actions/functions too so I kept it.

cbartz (dismissed):
LGTM

javierdelapuente (dismissed):
LGTM

weiiwang01 (dismissed):

docs/how-to/upgrade-wordpress-charm.md
```diff
@@ -0,0 +1,20 @@
+# How to upgrade WordPress Charm
+
+Before updating the Charm you need to backup the database using mysql charm's `create-backup` action.
```
erinecon, line 3: Notpick: mysql --> MySQL

docs/how-to/upgrade-wordpress-charm.md
```diff
@@ -0,0 +1,20 @@
...
+
+```
+juju run mysql/leader create-backup
+```
+Additional info can be found about backup on [Charmhub](https://charmhub.io/mysql/docs/h-create-and-list-backups)
```
erinecon, line 8: Nitpick: "on [Charmhub](link)" --> "in [the MySQL documentation](link)"
(To be more explicit about where the link takes the reader)

erinecon (approved):
LGTM

javierdelapuente (approved):

PR #366, opened 2026-06-02
Feat/jupyter notebook docs

yhaliaw:
# What this PR does

Converted the `docs/tutorial.md` to use Jupyter notebook.

Add configurations to cache the Jupyter notebook cell output of the `docs/tutorial.md`.

Add workflows for:
- Execute the `docs/tutorial.md` and save the cached Jupyter notebook output as an artifact in PRs.
- Tag the latest commit of a PR, and publish the artifact under it when PR is merged to main.

Add a custom script during the docs build process to download the cached output

# Why we need it

The idea is for the `docs/tutorial.md` to have real executed output.
However, charm deployment is a bit heavy for the docs buiid process on RTD.
So the execution happens in GitHub runners, and the docs build process on RTD only downloads the cache output and put it in the docs.

# Checklist

- [x] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [ ] I added or updated the documentation (if applicable)
- [ ] I updated `docs/changelog.md` with user-relevant changes
- [x] I used AI to assist with preparing this PR
- [ ] I added or updated tests as needed (unit and integration)
- [ ] **If integration test modules are used:** I updated the workflow configuration  
      (e.g., in `.github/workflows/integration_tests.yaml`, ensure the `modules` list is correct)
- [ ] **If this PR involves a Grafana dashboard:** I added a screenshot of the dashboard
- [ ] **If this PR involves Rockcraft:** I updated the version


docs/requirements.txt
```diff
@@ -33,3 +33,4 @@ vale==3.13.0.0
 
 # Additional extensions 
 sphinxcontrib-mermaid==2.0.0
+bash_kernel
```
yanksyoon: Should we pin this dependency?

yanksyoon (approved):
Very cool, thank you @yhaliaw

docs/requirements.txt
yhaliaw: Done.

docs/requirements.txt
```diff
@@ -2,7 +2,7 @@
 canonical-sphinx==0.5.2
 
 # Extensions previously auto-loaded by canonical-sphinx
-myst-parser==4.0.1
+myst-nb==1.1.2
```
erinecon: Just checking that `myst-nb` performs the same functions as `myst-parser` -- from what I understand, the starter pack uses `myst-parser` so that we can parse Markdown files and render them correctly on RTD. Does `myst-nb` do the same thing? As far as I can tell, `myst-nb` only cares about Jupyter notebooks.

The PR preview looks fine FWIW -- all the other pages still seem to render correctly.

yhaliaw: I have added `myst-parser` back as an explicit dependency.
I think it was still installed due to `my-parser` is a dependency of `my-nb`.

.github/workflows/execute_tutorial_notebook.yaml
erinecon, line 1: For my understanding: Why are there two separate workflows for the Jupyter notebooks? Why not combine them into a single workflow?

docs/_extensions/fetch_tutorial_notebook.py
erinecon, line 1: I'm wondering whether we should include this extension in wordpress-k8s-operator directly, or if we should move it into a dedicated repository for other projects to include and use. Otherwise we might duplicate this script across multiple projects.

What do you think?

docs/_extensions/fetch_tutorial_notebook.py
```diff
@@ -0,0 +1,65 @@
...
+
+import logging
+import os
+import urllib.request
+from pathlib import Path
+from urllib.error import URLError
+
+logger = logging.getLogger(__name__)
+
+_GITHUB_REPO = "canonical/wordpress-k8s-operator"
```
erinecon, line 14: If we want to eventually ship this extension as something that anyone could include in their Sphinx project, then we will need to find a way to not hardcode the repository in the Python script.

docs/conf.py
```diff
@@ -371,3 +380,21 @@
...
+    We work around this by making setup_myst_parser a no-op before loading myst_nb.
+    """
+    import myst_parser.sphinx_ext.main as myst_main
+
+    # canonical_sphinx already called setup_sphinx (via myst_parser), so all myst_parser
+    # roles, directives, transforms, and config values are already registered.
+    # Make it a no-op so myst_nb doesn't re-register them.
+    myst_main.setup_sphinx = lambda app, load_parser=False: None
+
+    app.setup_extension("myst_nb")
```
erinecon, line 400: I wonder whether there's a better way of incorporating your extension into canonical_sphinx, without needing to add this function to the conf.py file. My goal is to avoid having to add this to multiple projects if we want to expand this support to more of our team's tutorials.

CC @jahn-junior

erinecon (commented):
Thanks for working on adding support for Jupyter notebooks to our documentation 🎉 

A couple of questions to help my understanding and to anticipate potential risks (e.g. scalability and duplication)

.github/workflows/execute_tutorial_notebook.yaml
yhaliaw: One is run on pull request commits. The execution of the Jupyter notebook is done on each commit to guard against a commit with a failed execute Jupyter notebook workflow being merged to main.

The publish workflow is run on push to main, e.g., merge PR to main. Since it should only be published once it is in main.

docs/_extensions/fetch_tutorial_notebook.py
yhaliaw: I think it should eventually be moved to another repo, which all our repos use.
However, for the PoC I would need to test it more, before moving it.

docs/_extensions/fetch_tutorial_notebook.py
yhaliaw: Yes, I agree in the future it should.
However, the current code has not been tested with Read The Docs hosted yet. We can do that after this is merged to main.

.github/workflows/execute_tutorial_notebook.yaml
erinecon: Ahhhh okay, thank you for clarifying!

docs/_extensions/fetch_tutorial_notebook.py
erinecon: Okay cool, that sounds good. We can keep the script in this repo for development purposes.

PR #387, opened 2026-09-08
fix(ci): poll for execute-notebook run to fix docs-latest race

yhaliaw:
## Summary
- The push-triggered `Publish Tutorial Notebook` workflow looks up the merged PR's `Execute Tutorial Notebook` run (triggered separately by the `pull_request` event) and checked for a successful run exactly once.
- That execute run can still be in progress when the merge's `push` event fires — observed for PR #382 (commit `90d0cdb`), where the execute run took ~10 minutes but the publish job checked after only ~30 seconds and found nothing, so it silently skipped and left the `docs-latest` release stale (pointing at an older commit).
- This stale release is what `docs/_extensions/fetch_tutorial_notebook.py` serves to Read the Docs, and what `docs/_extensions/cache_notebook_from_artifact.py` (via `make serve-with-artifact`) ultimately depends on being fresh.

## Change
- `find-run` step now polls `gh run list` for up to 15 minutes (30s interval) until the execute run for the PR's head commit reaches `status=completed`, instead of checking once for `status=success`.

## Test plan
- [ ] Merge a docs-affecting PR and confirm `docs-latest`'s `commit.txt` updates to the merge commit even when `Execute Tutorial Notebook` is still running when `main` is pushed to.
- [ ] Confirm the workflow still exits gracefully (`found=false`) if the execute run ultimately fails or is not found within the timeout.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #388, opened 2026-09-08
fix(docs): local artifact fallback to docs-latest release

yhaliaw:
## Summary
- `Execute Tutorial Notebook` only triggers `on: pull_request`, so a checkout of a `main` merge commit never has a direct CI run for its own SHA. `make serve-with-artifact` / `make html-with-artifact` therefore failed unconditionally on `main` with "No successful Execute Tutorial Notebook run for commit ... yet" — verified against tip commit `90d0cdb`.
- `docs/_extensions/fetch_tutorial_notebook.py` already solves the same problem for Read the Docs by pulling the executed notebook from the `docs-latest` GitHub release (published by `Publish Tutorial Notebook`, see #387 for a related race fix in that workflow). The local script had no equivalent fallback.

## Change
- `cache_notebook_from_artifact.py` now falls back to downloading `docs-latest`'s `tutorial.ipynb` when there's no direct Execute Tutorial Notebook run for the current commit.
- Safety is unchanged: the existing code-cell match check in `_populate_cache` still rejects the notebook (with a clear error) if its code cells don't match the local `tutorial.ipynb`, so a stale release can't silently poison the cache.

## Test plan
- [x] Ran `make serve-with-artifact` on `main` tip (`90d0cdb`, no direct CI run) — fell back to the `docs-latest` release and built successfully.
- [ ] Run `make serve-with-artifact` from a PR branch commit that does have a direct `Execute Tutorial Notebook` run, to confirm the direct path still works unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yhaliaw:
Combined into #389.

